PR #3, opened 2020-06-12
feat: add sync and async esbuild transformers

danielroe:
Use these transformers like so:
```js
const jiti = require('jiti')(undefined, { transform: require('jiti/dist/esbuild-sync') })
const jiti = require('jiti')(undefined, { transform: require('jiti/dist/esbuild-async') })
```

A test suite has also been added using [`benchmark`](https://www.npmjs.com/package/benchmark):
```bash
# Tests init across esm js, babel js/ts, and esbuild js/ts
node ./test/bench/suite.js
# As above, but also tests several requires in a row
node ./test/bench/suite.js fixture
```


README.md
```diff
@@ -60,6 +60,13 @@ Cache directroy (only effective if `cache` is `true`)
 
 Transform function. See [src/babel](./src/babel.ts) for more details
 
+### `transformPreset`
+
+- Type: String
```
pi0, line 65: If adding a new option, what do you think of supporting `['esbuild', { async: false }]`?

index.js
```diff
@@ -5,7 +5,19 @@ module.exports = function (filename, opts) {
   opts = { ...opts }
 
   if (!opts.transform) {
-    opts.transform = require('./dist/babel')
+    switch (opts.transformPreset) {
```
pi0, line 8: We may also use a map object from name to callback. This way later on can do more refactors and support 3rd party presets

pi0: PS: It is currently not exactly preset but just transform impl. Does it makes sense having two options?

danielroe: Okay, let's revert to one.

PR #13, opened 2020-12-27
fix: remove duplicate v8-compile-cache

clarkdo:
v8-compile-cache has been imported from https://github.com/nuxt-contrib/jiti/blob/da34753d54e4bc726bb354dcbd77b4f3d7f7e0a0/lib/index.js#L9 , so I guess we can remove the it from jiti.js


pi0:
Thanks!

PR #15, opened 2021-01-20
feat: support node register hook

### LINKED ISSUE #14: jiti cli support
would be nice to be able to run a node script with `jiti ./script/index.ts` ...

danielroe:
support for `node -r jiti/register ./index.js`

closes  #14


lib/register.js
```diff
@@ -0,0 +1,11 @@
+const addHook = require('pirates').addHook
```
pi0, line 1:
```suggestion
const { addHook } = require('pirates')
```

lib/register.js
```diff
@@ -0,0 +1,11 @@
...
+
+const pwd = process.cwd()
+const jiti = require('.')(pwd)
```
pi0, line 4: passing nothing is similar to use pwd

lib/register.js
```diff
@@ -0,0 +1,11 @@
...
+
+addHook(
+  function (code, filename) {
```
pi0, line 7: Any reason not using arrow return? (also we can rename first param to source :P)

package.json
```diff
@@ -37,6 +37,7 @@
     "eslint": "latest",
     "esm": "latest",
     "mkdirp": "latest",
+    "pirates": "latest",
```
pi0, line 40: `lib/` is not bundled and adding to `dependencies`, makes jiti no longer zero dep. Maybe we can expose `register` from src/jiti (jiti instance)?

lib/register.js
danielroe: didn't want it to look too modern :stuck_out_tongue_winking_eye: - will use arrow

package.json
danielroe: Great idea. Will implement

pi0 (approved):
Fantastic!

package.json
```diff
@@ -7,6 +7,10 @@
   "main": "./lib/index.js",
   "types": "dist/jiti.d.ts",
   "bin": "bin/jiti.js",
+  "exports": {
+    "./": "./",
+    "./register": "./lib/register.js"
```
pi0, line 12: Last thing since still exports is not available everywhere, what if we move `register.js` to top level to make it working out of the box with `require('jiti/register')` everywhere?

danielroe: Good idea 👍

Issue #27, opened 2021-04-01
Add `@babel/plugin-proposal-decorators` to `transform`


LukeCSamuel-RV:
We make a lot of use of Typescript decorators in our code base.  In Nuxt 2.14 they could be used in places outside of webpack module resolution, such as Nuxt modules.  In Nuxt 2.15 they are no longer supported, because the babel instance bundled with Jiti doesn't have the `@babel/plugin-proposal-decorators` plugin.

I would propose that if `opts.ts` and ideally only when `experimentalDecorators === true` (from tsconfig), this plugin be added: `_opts.plugins!.push([require('@babel/plugin-proposal-decorators'), { legacy: true }])`

pi0:
Hi @LukeCSamuel-RV. Thanks for issue it makes sense adding for typescript. Would you please open a pull request?

tobiasdiez:
Could `babel-plugin-transform-typescript-metadata` also please be added so that the metadata added by these decorators can be used, say for dependency injection.

This would fix https://github.com/nuxt/typescript/issues/494.

PR #28, opened 2021-04-02
feat: support legacy decorators for typescript

LukeCSamuel-RV:
Addresses #27 

Feedback welcome


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jiti",
-  "version": "1.6.4",
+  "version": "1.7.0",
```
pi0, line 3: This number is auto increased by build script 
```suggestion
  "version": "1.6.4",
```

src/jiti.ts
```diff
@@ -80,6 +80,12 @@ export default function createJITI (_filename: string = process.cwd(), opts: JIT
     try { return nativeRequire.resolve(id, options) } catch (e) {}
   }
 
+  let tsConfig = { compilerOptions: {} as any }
+  const tsConfigPath = tryResolve('tsconfig')
```
pi0, line 84: Jiti, by design, acts independently of the environment to have predictive results. We cannot depend on tsconfig.json or babelrc

src/babel.ts
```diff
@@ -13,12 +13,17 @@ export default function transform (opts: TransformOptions): TRANSFORM_RESULT {
       [require('@babel/plugin-transform-modules-commonjs'), { allowTopLevelThis: true }],
       [require('babel-plugin-dynamic-import-node'), { noInterop: true }],
       [require('babel-plugin-transform-import-meta')],
-      [require('@babel/plugin-syntax-class-properties')]
+      [require('@babel/plugin-proposal-class-properties'), { loose: true }]
```
pi0, line 16: Is this change necessary for decorators' support? (if yes, please provide an example) Since modern environments already support class properties, we only add syntax support.

LukeCSamuel-RV: I was following Babel docs which recommended `proposal` over `syntax`, but these may be out of date.  Verified that `syntax` appears to be compatible with the decorators proposal, so I've switched it back.

pi0 (approved):
Thanks!

PR #29, opened 2021-04-06
feat: support legacy parameter decorators for typescript

LukeCSamuel-RV:
During further investigation, I discovered that the `@babel/plugin-proposal-decorators` plugin does not enable support decorators applied to method/constructor properties.  This PR is to enable support for these decorators as well.


pi0:
Hi @LukeCSamuel-RV thanks for PR. Would you please create a fixture in `test/fixtures/typescript/decoratos.ts` covering both usecases? (tests are locally only now)

LukeCSamuel-RV:
@pi0 I might need some guidance, running `node ./test/jiti.js` as shown in the README doesn't work as the file doesn't exist.  Do I need to add a file similar to the one under `bench/jiti.js` to initialize Jiti and test the typescript module?

pi0:
You can try it out via `yarn jiti ./test/path/to/file.ts` (a `yarn build` step is necessary). BTW sorry for lack of enough contrib docs. We need to have better testing.

LukeCSamuel-RV:
All good and thanks for the help, this is my first time contributing to OSS so I appreciate your quick responses!

pi0:
Thanks for PR :)

PR #31, opened 2021-05-09
fix: spread when pushing babel plugins from transformOptions

pimlie:
Without the spread the following error will be logged:
```
[jiti] Error: .plugins[4][1] must be an object, false, or undefined
```


danielroe (approved):

pi0:
fixed in v1.92.

PR #39, opened 2021-09-11
feat: add v8cache option

### LINKED ISSUE #38: errors related to v8-compile-cache
`jiti` require `v8-compile-cache` by default, but when importing esm in cjs context, it will throw an error.

---
```javascript
// index.js
require('v8-compile-cache');  // remove this line works fine
const lib = require('./lib');
lib();

// lib.js
module.exports = async function main() {
  const { default: esm } = await import('./esm.mjs');
  esm();
};

// esm.mjs
export default function () {
  console.log('esm');
}
```
TypeError [ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING]: A dynamic import callback was not specified.

https://github.com/windicss/windicss-webpack-plugin/issues/93

tmkx:
close #38

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,7 +10,7 @@ function onError (err) {
 }
 
 module.exports = function (filename, opts) {
-  if (!opts || opts.v8cache !== false) {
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for investigating issue and fix <3

PR #46, opened 2021-10-08
add ability to easily debug ts files

ast-pmykhailov:


acibilous:
Gues, could someone merge it?

pi0:
Hi @ast-pmykhailov. Would you please make a reprodution first when debugging ts files is not working? Normally jiti does not uglify or minify and preserves lines as well in error stacks.

acibilous:
@pi0 , the problem is not in uglify/minify of ts modules, but source maps for files are not available

pi0:
It is intentional. Jiti does not generate source maps but preserves line mapping for correct stack trace.

Would you please fork this repo to reproduce when mapping is not working for debugging: https://stackblitz.com/edit/node-tj6xqn?terminal=start

ast-pmykhailov:
@pi0 
we have .ts file we want to debug
image

On the first run, right after the build of jiti, it debugs as .js
image

On the second run, it works with .ts
But with this fix it works correct even for the first run right after the build

acibilous:
any updates @pi0 ? sorry for ping you, but it's really uncomfortable during debug(

pi0:
Hi, @acibilous Sorry for checking on this late. I see what you mean now. Added a few changes since it can be also useful for ESM files, option is now called `JITI_SOURCE_MAPS`.

PR #50, opened 2022-01-25
chore: setup tests

antfu:
This PR sets up snapshot testing with Vitest + Execa.

With this PR, it also surfaced an error in fixture `async` that dynamic import fail to resolve `.js` and fixed in b41e2755dc32f4454514e6d4e32d2ad7f946784e. Will make it work after we made the native esm support.


pi0 (approved):
Thanks 💯

PR #52, opened 2022-01-25
fix: support resolving with import condition

antfu:
We had the problem in Nuxt development that `jiti` failed to resolve `estree-walker`, which is due to it only having `exports` field without `main` (https://unpkg.com/browse/estree-walker@3.0.1/package.json).

Since we also transpile node modules, making the resolving aware of `exports` by `esm-resolve` package is able to fix most of the cases. This also makes the native ESM support a performance improvement instead of a blocker.


src/jiti.ts
```diff
@@ -1,9 +1,10 @@
 import { existsSync, readFileSync, writeFileSync } from 'fs'
 import { Module, builtinModules } from 'module'
-import { dirname, join, basename, extname } from 'path'
+import { dirname, join, basename, extname, resolve as pathResolve } from 'path'
 import { tmpdir } from 'os'
 import vm from 'vm'
 import { fileURLToPath } from 'url'
+import createESMResolver from 'esm-resolve'
```
pi0, line 7: Can we please use `mlly.resolvePathSync` ?

src/jiti.ts
```diff
@@ -77,12 +78,24 @@ export default function createJITI (_filename: string = process.cwd(), opts: JIT
     try { return nativeRequire.resolve(id, options) } catch (e) {}
   }
 
+  const _esmResolve = createESMResolver(_filename, { constraints: ['node'] })
```
pi0, line 81: Since we are still in cjs context, would suggest adding `require` constraint.

pi0:
> We had the problem in Nuxt development that jiti failed to resolve estree-walker, which is due to it only having exports field without main (https://unpkg.com/browse/estree-walker@3.0.1/package.json).

 Actually native `require.resolve` that is currently being used supports `exports` field but it uses `require` condition and `estree-walker` does not provide it.

src/jiti.ts
pi0: (as a workaround we can try/catch and resolve with `require` to hotfix estree issue but primarily prefer cjs to avoid unnecessary jiti transforms)

antfu:
Switched to mlly, but it does not seem to resolve the entry

pi0:
> Switched to mlly, but it does not seem to resolve the entry

We need to fallback to `['node', 'import']` condition to fix `estree-walker` issue :D

antfu:
Should be good now, thanks for the help :)

pi0 (approved):

pi0:
Because of side-effects to legacy users, moved this under a flag (https://github.com/unjs/jiti#esmresolve). Let's enable in next major version by default with more tests.

Issue #65, opened 2022-06-17
Detect native esm using `type:module`


pi0:
Currently, we detect if a module is native esm using by comparing extension against [.mjs](https://github.com/unjs/jiti/blob/912645967d095b7d1a97829e2013019b948a9761/src/jiti.ts#L226). We should also detect this by checking if closest `package.json` has `type:module`

PR #67, opened 2022-06-17
feat: detect native esm using `type: module`

danielroe:
resolves https://github.com/unjs/jiti/issues/65

### DIFF
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -236,7 +236,7 @@ export default function createJITI (_filename: string, opts: JITIOptions = {}, p
     )
 
     if (needsTranspile) {
-      debug(`[transpile${isNativeModule ? ' esm' : ''}]`, filename)
+      debug(`[transpile${isNativeModule ? ' esm module' : ''}]`, filename)
       source = transform({ filename, source, ts: isTypescript })
     } else {
       try {


src/jiti.ts
```diff
@@ -223,7 +223,20 @@ export default function createJITI (_filename: string, opts: JITIOptions = {}, p
 
     // Transpile
     const isTypescript = ext === '.ts'
-    const isNativeModule = ext === '.mjs'
+    let isNativeModule = ext === '.mjs'
+    let path = filename
+    if (ext === '.js') {
+      while (!isNativeModule && path && path !== '.' && path !== '/') {
```
pi0, line 229: Maybe we can extract this function to utils and also use pkg-types for finding closest package.json?

danielroe: `pkg-types` would have been my first choice but it's async

pi0: Shall we introduce a `sync` method there? It might be useful for edge cases like this and scripts.

src/jiti.ts
```diff
@@ -235,7 +236,7 @@ export default function createJITI (_filename: string, opts: JITIOptions = {}, p
     )
 
     if (needsTranspile) {
-      debug('[transpile]', filename)
+      debug(`[transpile${isNativeModule ? ' native' : ''}]`, filename)
```
pi0, line 239: CJS is also native :) 

```suggestion
      debug(`[transpile${isNativeModule ? ' esm' : ''}]`, filename)
```

pi0: module is better

pi0: esm also stands for "EcmaScript Modules". However, i understand it can also be misleading considering non standard ESM syntax but then I would call that "bad esm" :D

src/jiti.ts
```diff
@@ -235,7 +236,7 @@ export default function createJITI (_filename: string, opts: JITIOptions = {}, p
     )
 
     if (needsTranspile) {
-      debug('[transpile]', filename)
+      debug(`[transpile${isNativeModule ? ' esm' : ''}]`, filename)
```
pi0, line 239:
```suggestion
      debug(`[transpile${isNativeModule ? ' esm module' : ''}]`, filename)
```

PR #69, opened 2022-06-20
chore: migrate to pnpm

danielroe:


pi0:
@danielroe It seems types are broken with the new inline plugin. Can you please check this?

pi0 (approved):
Thanks

Issue #106, opened 2022-12-29
Release for `satisfies` operator support


sxzz:
### Environment

N/A

### Reproduction

N/A

### Describe the bug

Please bump a new version for `satisfies` operator support

It's already supported via https://github.com/unjs/jiti/commit/e79ba40f3db92d250228cb14434d550cbafe88e1

pi0:
Released 1.16.1. I hope it adds support 👍🏼  (PR welcome to add fixture)

sabinmarcu:
Thank you, @pi0 , fixed the issue for me 🙏 

Made a PR for adding a fixture for the keyword. Hope it helps (let me know if I need to do anything other than what's already in the PR)

PR #107, opened 2023-01-03
test: add typescript satisfies fixture

sabinmarcu:
Adding a fixture for the `satisfies` keyword in typescript (as surfaced in https://github.com/unjs/jiti/issues/106). Hopefully it's good enough 😅 

Please let me know if there's anything else I should be doing to help 👍


pi0 (approved):
💯 Thanks

PR #110, opened 2023-01-10
Added Example for debugging with Visual Studio Code and source maps

Hecatron:
Hi,
It took me a little while to figure out the right combination of options to use when debugging typescript using visual studio code and jiti.
I figured I'd add an example just to make it easier for others.

I noticed that when the cache is enabled, the cached files are not overwritten if the sourcemap option was disabled the first time around but then enabled the second time around when launching a script.
Also disabling the cache seems to be a necessity anyway when debugging with source maps, as I think the sourcemap changes on each run.

### DIFF
--- a/examples/vscode-debug/.vscode/launch.json
+++ b/examples/vscode-debug/.vscode/launch.json
@@ -1,0 +1,29 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug Typescript",
+      "request": "launch",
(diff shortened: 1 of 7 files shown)


daxaxelrod:
Holy cow, I hope you have an amazing day today, you certainly saved mine.

pi0:
I appreciate your work on this. With introduction of `jiti/register`, i think we can instead suggest people to directly use Node.js ESM loader for better debugging.

PR #112, opened 2023-01-30
feat: add support for `.mts`

webstrand:
Without these changes, jiti is treating `.mts` files as non-modules. Probably should add support for `.cts`, too, but I don't have a test-case for it.

### DIFF
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -43,7 +43,7 @@ const defaults: JITIOptions = {
   esmResolve: _EnvESMResolve || false,
   cacheVersion: "7",
   legacy: lt(process.version || "0.0.0", "14.0.0"),
-  extensions: [".js", ".mjs", ".cjs", ".ts", ".mts"],
+  extensions: [".js", ".mjs", ".cjs", ".ts", ".mts", ".cts"],
   alias: _EnvAlias,
   nativeModules: _EnvNative || [],
   transformModules: _EnvTransform || [],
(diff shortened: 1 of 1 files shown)


src/jiti.ts
```diff
@@ -43,7 +43,7 @@ const defaults: JITIOptions = {
   esmResolve: _EnvESMResolve || false,
   cacheVersion: "7",
   legacy: lt(process.version || "0.0.0", "14.0.0"),
-  extensions: [".js", ".mjs", ".cjs", ".ts"],
+  extensions: [".js", ".mjs", ".cjs", ".ts", ".mts"],
```
pi0, line 46:
```suggestion
  extensions: [".js", ".mjs", ".cjs", ".ts", ".mts", ".cts"],
```

src/jiti.ts
```diff
@@ -286,9 +286,10 @@ export default function createJITI(
     let source = readFileSync(filename, "utf8");
 
     // Transpile
-    const isTypescript = ext === ".ts";
+    const isTypescript = ext === ".ts" || ext === ".mts";
```
pi0, line 289:
```suggestion
    const isTypescript = ext === ".ts" || ext === ".mts" || ext === ".cts";
```

src/jiti.ts
```diff
@@ -286,9 +286,10 @@ export default function createJITI(
...
     const isNativeModule =
       ext === ".mjs" ||
+      ext === ".mts" ||
```
pi0, line 292: Since node.js has not native support for `.mts`, i think we shouldn't have it here

pi0 (approved):
Thanks!

pi0:
I will release it shortly this week (checking some other minor issue before).

PR #123, opened 2023-02-28
fix: workaround for pnpm and `TMPDIR`

### LINKED ISSUE #120: fix(docker): `node-jiti` directory created in project root
### Environment

Docker
jiti 1.17.1
node 19.7.0-alpine

### Reproduction

https://github.com/dargmuesli/nuxt-jiti

### Describe the bug

1. `docker build -t nuxt-jiti .`
2. `docker run --rm -it nuxt-jiti /bin/sh`
3. `ls -la`
4. see that **there is a `node-jiti` directory in the project's root**

I'd expect the directory to be under `/tmp`.

### Additional context

1. remove `"postinstall": "nuxt prepare"` from the `package.json`
2. append `RUN pnpm nuxt prepare` to the `Dockerfile`
3. run all steps above
4. **see that `node-jiti` is created in `/tmp`**, as expected

dargmuesli:
Resolves #120

Workaround for pnpm setting an incorrect `TMPDIR`. See https://github.com/pnpm/pnpm/issues/6140

Thinking about it a bit more I worry that introducing this workaround also prevents people from making use of their custom `TMPDIR` - if there is one. I think that more people run `nuxt prepare` using pnpm in the respective hook, though, than people set `TMPDIR` themself. So I also added that `JITI_TMPDIR_KEEP` can be set to a truthy value to disable this workaround for people who'd like to keep their `TMPDIR`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,14 +5,18 @@ import { join } from "pathe";
 import type { PackageJson } from "pkg-types";
 
 export function getCacheDir() {
-  let _tmpDir = tmpdir()
+  let _tmpDir = tmpdir();
 
   // Workaround for pnpm setting an incorrect `TMPDIR`.
   // Set `JITI_RESPECT_TMPDIR_ENV` to a truthy value to disable this workaround.
   // https://github.com/pnpm/pnpm/issues/6140
   // https://github.com/unjs/jiti/issues/120
-  if (process.env.TMPDIR && _tmpDir === process.cwd() && !process.env.JITI_RESPECT_TMPDIR_ENV) {
-    const _env = process.env.TMPDIR
+  if (
+    process.env.TMPDIR &&
+    _tmpDir === process.cwd() &&
+    !process.env.JITI_RESPECT_TMPDIR_ENV
+  ) {
+    const _env = process.env.TMPDIR;
     delete process.env.TMPDIR;
     _tmpDir = tmpdir();
     process.env.TMPDIR = _env;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -8,16 +8,16 @@ import type { PackageJson } from "pkg-types";
 // Set `JITI_TMPDIR_KEEP` to a truthy value to disable this workaround.
 // https://github.com/pnpm/pnpm/issues/6140
 // https://github.com/unjs/jiti/issues/120
-export function getOsTmpDir() {
+export function getCacheDir() {
   if (process.env.JITI_TMPDIR_KEEP) {
-    return tmpdir();
+    return join(tmpdir(), "node-jiti");
   }
 
   const currentTmpDir = process.env.TMPDIR;
   delete process.env.TMPDIR;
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -1,8 +1,25 @@
...
+// Set `JITI_TMPDIR_KEEP` to a truthy value to disable this workaround.
+// https://github.com/pnpm/pnpm/issues/6140
+// https://github.com/unjs/jiti/issues/120
+export function getCacheDir() {
+  if (process.env.JITI_TMPDIR_KEEP) {
+    return join(tmpdir(), "node-jiti");
+  }
+
+  const currentTmpDir = process.env.TMPDIR;
+  delete process.env.TMPDIR;
```
pi0, line 17: If I'm current this situation only happens when `TMPDIR === cwd()` right? (otherwise we can support normal use of `TMPDIR`

src/utils.ts
```diff
@@ -1,8 +1,25 @@
...
+import { tmpdir } from "os";
 import { join } from "pathe";
 import type { PackageJson } from "pkg-types";
 
+// Workaround for pnpm setting an incorrect `TMPDIR`.
+// Set `JITI_TMPDIR_KEEP` to a truthy value to disable this workaround.
+// https://github.com/pnpm/pnpm/issues/6140
+// https://github.com/unjs/jiti/issues/120
+export function getCacheDir() {
+  if (process.env.JITI_TMPDIR_KEEP) {
```
pi0, line 12: We could support env always as `JITI_CACHE_DIR` :D

src/utils.ts
```diff
@@ -1,8 +1,25 @@
 import { lstatSync, accessSync, constants, readFileSync } from "fs";
 import { createHash } from "crypto";
+import { tmpdir } from "os";
 import { join } from "pathe";
 import type { PackageJson } from "pkg-types";
 
+// Workaround for pnpm setting an incorrect `TMPDIR`.
+// Set `JITI_TMPDIR_KEEP` to a truthy value to disable this workaround.
+// https://github.com/pnpm/pnpm/issues/6140
+// https://github.com/unjs/jiti/issues/120
```
pi0, line 10: Refactor: i think we can move comments re workaround above the relevant lines

src/utils.ts
dargmuesli: I think that could be a separate feature to this workaround, because replaceing `JITI_TMPDIR_KEEP` with `JITI_CACHE_DIR` would require users to give their cache dir twice (once `TMPDIR` and once `JITI_CACHE_DIR`) if they'd want to work around the pnpm issue. I'd favor providing the workaround by default and allowing to fallback to a custom `TMPDIR` with a toggle.
Again, I understand that a `JITI_CACHE_DIR` sounds good, but can be implemented separately.

dargmuesli:
I've added tests for the new utility function. I'm new to `vitest`, I hope the tests are well written :wink: I'm always open for feedback.

PR #128, opened 2023-03-14
feat: allow resolving `.ts` files with `.js` extension

### LINKED ISSUE #96: Typescript imports with ".js" file extension are failing.
We should support importing `.ts` files via `.js`. 

Related issue:
https://github.com/vitejs/vite/issues/6671

Also, we should support `.mts` and `.cts` by default.

Akryum:
fix #96

### DIFF
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -175,7 +175,7 @@ export default function createJITI(
     for (const ext of _additionalExts) {
       resolved =
         tryResolve(id + ext, options) ||
-        tryResolve(id.replace(/(c|m)?j(sx?)$/, "$1t$2"), options) ||
+        tryResolve(id.replace(/\.(c|m)?j(sx?)$/, ".$1t$2"), options) ||
         tryResolve(id + "/index" + ext, options);
       if (resolved) {
         return resolved;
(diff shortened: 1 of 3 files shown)


src/jiti.ts
```diff
@@ -175,7 +175,7 @@ export default function createJITI(
     for (const ext of _additionalExts) {
       resolved =
         tryResolve(id + ext, options) ||
-        tryResolve(id.replace(/(c|m)?j(sx?)/, "$1t$2"), options) ||
+        tryResolve(id.replace(/(c|m)?j(sx?)$/, "$1t$2"), options) ||
```
pi0, line 178: Can regex be more specific with dot ext?

Akryum: Done

src/jiti.ts
```diff
@@ -178,6 +175,7 @@ export default function createJITI(
     for (const ext of _additionalExts) {
       resolved =
         tryResolve(id + ext, options) ||
+        tryResolve(id.replace(/\.(c|m)?j(sx?)$/, ".$1t$2"), options) ||
```
pi0, line 178:
```suggestion
        // TODO: Only do this for imports within a typescript parent module
        tryResolve(id.replace(/\.(c|m)?j(sx?)$/, ".$1t$2"), options) ||
```

pi0 (approved):
Thanks 💯

PR #129, opened 2023-03-15
feat: support `import.meta.env`

### LINKED ISSUE #127: Support `import.meta.env`
https://github.com/unjs/jiti/issues/101#issuecomment-1465363353

pi0:
resolves #127. I have used a fork of `@import-meta-env/babel` as it was depending on dotenv library. Also we needed simple `process.env` accessor. would be happy to move to upstream if exports similar transform options.


src/babel.ts
```diff
@@ -54,6 +59,7 @@ export default function transform(opts: TransformOptions): TRANSFORM_RESULT {
   }
 
   try {
+    console.log(transformSync(opts.source, _opts)?.code);
```
danielroe, line 62:
```suggestion
```

PR #134, opened 2023-04-04
fix: resolve Windows paths properly

userquin:
When stubbing (unbuild), Windows paths are absolute but there is no prototol, and so will have unit (`C://...`), adding Windows check and a regexp should fix the problem:

imagen

### DIFF
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -269,7 +269,7 @@ export default function createJITI(
     } else if (id.startsWith("file:")) {
       id = fileURLToPath(id);
     } else if (isWindows && /^w+:/.test(id)) {
-      id = `file:///${id.replace(/\\/g, "/")}`;
+      id = fileURLToPath(`file:///${id.replace(/\\/g, "/")}`);
     }
 
     // Check for builtin node module like fs
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -269,7 +269,7 @@ export default function createJITI(
     } else if (id.startsWith("file:")) {
       id = fileURLToPath(id);
     } else if (isWindows && /^w+:/.test(id)) {
-      id = `file:///${id.replace(/\//g, "\\")}`;
+      id = `file:///${id.replace(/\\/g, "/")}`;
     }
 
     // Check for builtin node module like fs
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -268,7 +268,7 @@ export default function createJITI(
       id = id.slice(5);
     } else if (id.startsWith("file:")) {
       id = fileURLToPath(id);
-    } else if (isWindows && /^w+:/.test(id)) {
+    } else if (isWindows && /^\w:/.test(id)) {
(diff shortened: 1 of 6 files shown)


src/jiti.ts
```diff
@@ -268,6 +268,8 @@ export default function createJITI(
       id = id.slice(5);
     } else if (id.startsWith("file:")) {
       id = fileURLToPath(id);
+    } else if (isWindows && /^w+:/.test(id)) {
```
danielroe, line 271:
```suggestion
    } else if (isWindows && /^\w:/.test(id)) {
```

src/jiti.ts
```diff
@@ -268,6 +268,8 @@ export default function createJITI(
...
+      id = `file:///${id.replace(/\\/g, "/")}`;
```
danielroe, line 272: we could use `url.pathToFileURL`

userquin:
cannot build project, I'll fix also Windows on the repo, adding rimraf to cleanup folders (rn using `rm`)...

src/jiti.ts
userquin: yeah, I'm fixing it rn, thx Daniel

userquin: done

src/jiti.ts
```diff
@@ -268,6 +268,8 @@ export default function createJITI(
       id = id.slice(5);
     } else if (id.startsWith("file:")) {
       id = fileURLToPath(id);
+    } else if (isWindows && /^\w:/.test(id)) {
+      id = fileURLToPath(`file:///${id.replace(/\\/g, "/")}`);
```
danielroe, line 272: This will return `/C:/test` - do we want `pathToFileURL` instead?

userquin: `/` is a valid path, the problem on Windows is `<unit>:` is not a valid protocol, `fs` variants should handle `/<unit>:...` properl: we're just adding `file:///` prefix, and so it will be handle in the same way in previous `else if` statement.

FYI: running tests without this PR, all failing:

imagen

With this PR, all test ok:

imagen

danielroe: would you add a test that would have failed without this change?

userquin: I'll try to figure out how to add it :thinking:

userquin: since jiti is using webpack, can we add some code cleanup on the build process? I can't think of anything to include the test, except to include a plugin (vite) that transforms jiti module in the test configuration

userquin: done

userquin:
closing this, it works, I'll send a new one PR to fix Windows build in the repo (use rimraf instead `rm`)

sadeghbarati:
@userquin 

I had same issue in macOS too, is unbuild stub working for you? you had normal or monorepo project?

I add your patch it's not working 😭

https://github.com/sadeghbarati/jiti-resolve-issue

userquin:
@sadeghbarati it's not working, that's why I close this PR, I need to check it again to find the root cause...

PR #141, opened 2023-05-23
test: `requireCache: false` casing infinite loop importing `lodash-es`

antfu:
Originally reported at https://github.com/unocss/unocss/issues/2651

**This PR adds a failing test as repro**

I haven't look deep, but a few observations:

- When `requireCache: false`, it casing infinite loop importing `lodash-es`
- Even when `requireCache: true`, it will take very long time (2s) to import `lodash-es`


danielroe:
Also linking https://github.com/nuxt/nuxt/issues/14443 and https://github.com/nuxt/nuxt/issues/15179.

pi0:
Thanks for the reproduction. https://github.com/unjs/jiti/pull/125 solved similar issues with recursive deps when require cache is disabled but seems we need supporting nested circular deps as well...

Re performance, it is because of actual (unnecessary) transform time. I suggest not to disable `cache` option. Filesystem cache will be automatically invalidated when needed.

PR #142, opened 2023-05-24
fix: handle nested circular dependencies

pi0:
Resolves #141

An extension of #125 to keep a shared cache from (main) parent and children while evaluating. This solves handling nested circular dependencie


pi0:
Thanks for reminder @antfu :

PR #156, opened 2023-08-09
feat: experimental native bun support

### LINKED ISSUE #132: missing `JITIOptions` type export
workaround

```ts
import jiti from "jiti";

type JITIOptions = Parameters<typeof jiti>[1];
```

### LINKED ISSUE #154: Bun compatibility/bypass

pi0:
### 🔗 Linked issue

resolves #154

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

When running in [Bun runtime](https://bun.sh/) we can leverage Bun's native typescript and CJS handling.

Running current fixtures against Bun, it seems platform support is extremely good and we can directly depend on Bun's `require` implementation 💯 

There might be more runtime differences but I think generally for the sake of performance, worth to make Jiti bypass all transforms. 

Known issues:
- `import.meta.env` is not supported
- Inside typescript files, cannot import with `.mts` as `.mjs`

On known issues, jiti falls back to it's default behavior to increase compatibility

/cc @birkskyum @paperdave @Jarred-Sumner

TODO:
- [x] Cache invalidation support (important for config reload like `nuxt.config.ts`)
- [x] Disable when custom transform options/plugins are provided (untyped babel plugin)
- [x] Allow opting-out (using `JITI_EXPERIMENTAL_BUN: false` or `experimentalBun: false`)

### 📝 Checklist

- [ ] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.


package.json
```diff
@@ -21,7 +21,8 @@
     "lint": "eslint --ext .ts,.js . && prettier -c src lib test stubs",
     "lint:fix": "eslint --fix --ext .ts,.js . && prettier -w src lib test stubs",
     "release": "pnpm build && pnpm test && changelogen --release --push && npm publish",
-    "test": "pnpm lint && vitest run --coverage"
+    "test": "pnpm lint && vitest run --coverage",
+    "test:bun": "JITI_DEBUG=1 bun test/bun/index.ts"
```
birkskyum, line 25: The `bun --bun` flag could be added here to make sure we're using bun througout.

pi0: Thanks for the tip. just did :)

Jarred-Sumner:
> Inside typescript files, cannot import with .mjs extension

This sounds like a bug. Does it really not work in general with .mjs or is there something more specific going wrong?

Jarred-Sumner:
To make import.meta.env work, I think you could override the getter on the import.meta object and define an env property to return either Bun.env or process.env. 

Something like 

```
const proto = Object.getPrototypeOf(import.meta)
proto.env = process.env 
```

pi0:
> To make import.meta.env work, I think you could override the getter on the import.meta object and define an env property to return either Bun.env or process.env.

Thanks! I wasn't sure if extending `import.meta` is something safe for Bun but i can add polyfill like this if seems good to you!

src/jiti.ts
```diff
@@ -49,6 +50,7 @@ const defaults: JITIOptions = {
   alias: _EnvAlias,
   nativeModules: _EnvNative || [],
   transformModules: _EnvTransform || [],
+  experimentalBun: _ExpBun === undefined ? !!process.versions.bun : !!_ExpBun,
```
AndreyYolkin, line 53: What do you think about moving Bun detection into https://github.com/unjs/std-env?

PR #158, opened 2023-08-16
feat: bypass transformation in browser and vitest

antfu:
Combining with https://github.com/unjs/unbuild/pull/300, it would allow `unbuild --stub`'s output compatible with the browser and Vitest, by opting-out jiti transform on them and forward to Vite's transformation.

Playground: https://stackblitz.com/~/github.com/antfu/testing-jiti-on-vite

### DIFF
--- a/lib/index.mjs
+++ b/lib/index.mjs
@@ -1,14 +1,13 @@
 import { isTest } from "std-env";
 
-let createRequire = () => {
-  throw new Error(
-    "createRequire is not supported in browser or test environment.",
-  );
-};
-
-try {
-  createRequire = await import("module").then((m) => m.createRequire);
-} catch {}
+const createRequire = await import("node:module")
+  .then((m) => m.createRequire)
+  // eslint-disable-next-line unicorn/consistent-function-scoping
+  .catch(() => () => {
+    throw new Error(
+      "createRequire is not supported in browser or test environment.",
+    );
+  });
 
 export default function (filename, opts) {
   const isBrowser =


pi0:
(drafting until we support `jiti.import` to avoid breaking changes for current jiti users)

antfu:
@pi0 Implemented the async `jiti.import` (https://github.com/unjs/jiti/pull/158/commits/eddf1f582ec1b7acfe234b6280b1c540f5172477) and let unbuild's mjs stub using it (https://github.com/unjs/unbuild/pull/300/commits/f3d5a566c844bfc4a9d504d5f4857a7cc2d7bfd3) for vite/vitest env. WDYT?

pi0:
Seems good start only would you mind separating PR for `jiti.import`?

lib/index.mjs
```diff
@@ -0,0 +1,35 @@
...
+  createRequire = await import("module").then((m) => m.createRequire);
+} catch {}
+
+export default function (filename, opts) {
+  const isBrowser =
+    typeof window !== "undefined" &&
+    typeof document !== "undefined" &&
+    typeof navigator !== "undefined";
+  const isVitest =
+    typeof process !== "undefined" && process.env && process.env.VITEST;
```
pi0, line 17: We can use `std-env` also would `isTest` be enough to make implementation generic?

lib/index.mjs
```diff
@@ -0,0 +1,35 @@
+let createRequire = () => {
+  throw new Error(
+    "createRequire is not supported in browser of test environment.",
+  );
+};
+
+try {
+  createRequire = await import("module").then((m) => m.createRequire);
```
pi0, line 8:
```suggestion
  createRequire = await import("node:module").then((m) => m.createRequire);
```

We can also add fallback in `.cache`

antfu: What do you mean by `.cache`?

lib/index.mjs
antfu: Not sure about other testing env, but generally we do this cause Vitest is based on Vite, let me test a bit with Jest at least. Maybe also for Bun?

pi0: Yeah. In general, if we can do it runtime-agnostic even if aiming for vitest for now, i would prefer that. If vite/vitest has special behavior, would be happy to also add specific condition (ideally flag from std-env) -- if we have to.

antfu: Jest seems to neve pick the ESM version, so yeah it should be fine I guess? Updated.

lib/index.mjs
```diff
@@ -0,0 +1,35 @@
...
+
+export default function (filename, opts) {
+  const isBrowser =
+    typeof window !== "undefined" &&
+    typeof document !== "undefined" &&
+    typeof navigator !== "undefined";
+
+  // In browsers, or in Vitest,
+  // we bypass jiti transform and fallback to the native import and let the runtime handles it
+  if (isBrowser || isTest) {
```
pi0, line 21: We cannot always assume this for browser envs. In an environment like jsdom, isBrowser can be true but also jiti could be working.

package.json
```diff
@@ -5,6 +5,7 @@
   "repository": "unjs/jiti",
   "license": "MIT",
   "main": "./lib/index.js",
+  "module": "./lib/index.mjs",
```
pi0, line 8: Currently, little bit hesitated to introduce it as `index.mjs` which is dedicated to fallback (test, browser) environments. It makes behavior partial and less consistent + the fact  that `module` is not a standard format and (test, browser) builders can also pick `exports` field (which we should add) and it can again break the functionality of what we add.

lib/index.mjs
pi0: I meaning chaining promise with catch! Will add code refactor.

lib/index.mjs
antfu: Yeah true, but I am not sure if there is a good way to distinguish jsdom and real browser. Any ideas?

package.json
antfu: Yeah, but not sure how we could do it otherwise (Vite throws when importing CJS). I guess maybe we could release a patch with `jiti.import`, update `unbuild` with `await jiti.import()`. And then have this as a major-beta. So users could opt-in this by override with the beta version of jiti.

I know quite a few ppl waiting for this capability for both frontend and tests

pi0: Seems a good plan for unbuild. I already wanted to start these as new major version.

Issue #32, opened 2021-05-27
native esm loader with `jiti.import`


pi0:
In contexts where file extension is explicit `.mjs` or closest `package.json` has `type: module` we may use native ESM and for typescript, disable CommonJS transforms. 

Blocker is, jiti is supposed to be used in a cjs context (require replacement). We probably need to implement a new function like `jiti.import` that returns Promise (compatible to be dynamic import replacement)

dioxine:
oh yeah, waiting this so much, cause `vite-plugin-windicss` uses `jiti` for loading configurations and thats why it cannot work with `type: module` `sveltejs/kit` exclusively uses.

pi0:
@danielroe I've moved vite 3 related issue to #66. Supporting native esm loader can probably wait until Node.js has it out of the flag.

dohooo:
Hi, Thanks for your effort! This project is very useful! 🥳

Uhhh, How to disable CommonJS transforms? Any workarounds are there?

pi0:
Hi. Just a quick update, i still plan to do this. In the meantime, are they are incomaptbile libraries with cjs transforms? Please mention them there might be faster fixes for them.

PR #170, opened 2023-10-02
feat: add `jiti.import` function for async import

antfu:
related to #32

It should be the very first step to have an async import function for ESM, which we could improve the implemtation later. Unblocks #158


pi0 (approved):

pi0:
Thanks! Will follow up with small refactors in next steps.

Issue #138, opened 2023-04-21
Support `.jsx` and `.tsx` by default?


tessro:
### Describe the feature

Would you be open to adding `.jsx` and `.tsx` extensions to the default set? Currently, in React projects, Jiti throws "Cannot find module" errors when parsing files that import JSX/TSX. TypeScript typically loads these automatically, so it's a cryptic issue that is likely to trip people up. (I spent over an hour tracking it down.) Since Jiti is used as middleware in other tools (in my case, by GraphQL Code Generator), this error is typically not something a user is prepared to debug, since they may not even realize Jiti is used under the hood.

JSX/TSX are typically supported automatically by TypeScript itself so this feels like a safe change, if you're open to it! 🤞🏼 

Addendum: For `.tsx` specifically, I believe `isTypescript` would also need to be updated for this to work. That feels extremely safe since `.tsx` files are always TypeScript.

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
I am open to supporting tsx. Do you mind to open a PR with change + a test fixture for usage? (how are we going to provide a default `h` function?)

paulrosania:
Sure, I will take a look at this. I'm not sure what you mean by `h` function, is this an additional functionality beyond the default import?

reichhartd:
@paulrosania Did you found a workaround until the feature is implemented?

paulrosania:
Unfortunately no, I have not had time to look into this further. For now I'm using this patch (via `patch-package`) which adds `.tsx` support. It's brittle but it gets the job done.

https://gist.github.com/paulrosania/4c21f6592e241e49d34f1fdc6586ef96

pi0:
`h` and `Fragment` are global constructs that a transpiled JSX code depends on.

For example this:

```tsx
export const Test = () => <div>42</div>;
```

It needs to be transpiled to this:

```js
export const Test = () => h("div", null, "42");
```

And jiti should support an option to specific what global `h` is (it can be `React.createElement` or `Vue.h` for example)

yuusheng:
I would like to solve this!

PR #200, opened 2024-01-05
feat: support jsx

yuusheng:
### 🔗 Linked issue

https://github.com/unjs/jiti/issues/138

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Currently `jiti` has no jsx support. This PR aims to support jsx with some simple configurations.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.


src/options.ts
```diff
@@ -23,11 +24,23 @@ const jitiDefaults: JITIOptions = {
...
+    ".mts",
+    ".cts",
+    ".json",
+    ...(_EnvJsx ? [".jsx", ".tsx"] : []),
+  ],
   alias: _EnvAlias,
   nativeModules: _EnvNative || [],
   transformModules: _EnvTransform || [],
   experimentalBun: _ExpBun === undefined ? !!process.versions.bun : !!_ExpBun,
+  transformOptions: {
```
yuusheng, line 41: This could be overwritten by user config. Should we use `defu` to merge the `jitiDefaults` and `userOptions`.

pi0: For this PR, we can simply spread user options. We can consider defu in later PR 👍🏼

src/babel.ts
```diff
@@ -44,6 +44,19 @@ export default function transform(opts: TransformOptions): TRANSFORM_RESULT {
     _opts.plugins.push(require("@babel/plugin-syntax-import-assertions"));
   }
 
+  if (opts.jsx) {
+    if (!_opts.presets) {
+      _opts.presets = [];
+    }
+
+    _opts.presets.push([
+      require("@babel/preset-react"),
```
pi0, line 53: I think we should base on `synctax-jsx` directly for basic/universal support https://babeljs.io/docs/babel-plugin-syntax-jsx

README.md
```diff
@@ -116,6 +117,14 @@ Return the `.default` export of a module at the top-level.
 
 Custom alias map used to resolve ids.
 
+### `jsx`
+
+- Type: Boolean
```
pi0, line 122: (we can document value can be a stringified JSON object too)

src/babel.ts
yuusheng: I use `@babel/preset-react` because `@babel/plugin-syntax-jsx` can only support babel to parse this syntax, but cannot transform to code to `h` function or `React.createElement`. Or we can use both  `@babel/preset-react` and `@babel/plugin-transform-react-jsx` to parse it and transform to `h` function (or `React.createElement`). `@babel/preset-react` just do it inside.

README.md
yuusheng: It has been updated.

README.md
```diff
@@ -14,6 +14,7 @@ Runtime Typescript and ESM support for Node.js
 
 - Seamless typescript and ESM syntax support
 - Seamless interoperability between ESM and CommonJS
+- `jsx` support with simple configuration
```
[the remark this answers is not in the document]
yuusheng: It has been updated.

src/babel.ts
yuusheng: I use plugins to replace presets.

yuusheng:
Should this PR merge into the `main` branch? Maybe merge into the `v1` branch?

src/babel.ts
pi0: Thanks for change. I will try to locally test but as explained in #138, my main idea to have minimal syntax changes and instead _inject_ the react (or any other JSX renderer) render function in instead of tightly coupling with react (although rest of the ecosystem for many of historical reasons are currently are)

pi0:
`main` should we fine we shall announce it for v2 👍🏼

src/babel.ts
yuusheng: I don't know if I understand correctly. It seems that we didn't tightly couple with react. Installing `react` as devDependencies is to solve the test error when running jsx files. And `@babel/plugin-transform-react-jsx` can transform code from `<div>foo</div>` to `React.createElement('div', null, 'foo')` (or `h('div', null, 'foo')` depends on configuration. That's all we do. And the react (or other JSX renderer) is injected by the users.

PR #239, opened 2024-06-05
feat: basic top-level await support

pi0:
This PR supports basic [top level await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#top_level_await) support via `jiti.import()` and `jiti <path>` CLI.

Implementation is fairly basic and adds support by wrapping the module into an `async function` instead of `function` that Node.js does.

In order to allow chained usage, since jiti transforms all imports to require internally, we replace ` = require(` patterns to ` = await require` in async contexts to make sure they are resolved. (this is more of a workaround but better than not having it!)

### DIFF
--- a/bin/jiti.js
+++ b/bin/jiti.js
@@ -5,7 +5,6 @@ const { resolve } = require("node:path");
 const script = process.argv.splice(2, 1)[0];
 
 if (!script) {
-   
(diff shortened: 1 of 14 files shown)


aryaemami59:
This is very exciting.

PR #250, opened 2024-06-13
docs: add detailed explanation of the function

RSS1102:
**Reference:** #
Cite [antfu's article](https://antfu.me/posts/publish-esm-and-cjs#stubbing) to provide a detailed explanation of its role.

Associated
https://github.com/unjs/unbuild/issues/398 
https://github.com/unjs/unbuild/pull/401


pi0:
Thanks for PR dear @RSS1102. It is correct info but it is in context of how unbuild works (i will check unbuild PR as soon as could)

PR #261, opened 2024-06-27
chore: add `unplugin-purge-polyfills`

danielroe:
This is an experiment with using unplugin-purge-polyfills in a webpack-based project.

It does not currently remove any polyfills - investigating.


pi0:
Do you still want to continue this experiment or if there are not any matches maybe closing it?

danielroe:
yes, still no matches - I can always reopen if we spot something in future.

PR #271, opened 2024-07-04
refactor: improve internal babel types

aryaemami59:
## **This PR**:

- [X] Provides missing type definitions for some of the babel related packages through module augmentations and ambient declaration files.
- [X] Corrects some of the extisiting types.
- [X] Fixes a very minor issue related to snapshots on Windows.


src/plugins/transform-module/dynamic-import.ts
```diff
@@ -23,7 +21,7 @@ const requireInterop = (
   return t.callExpression(t.memberExpression(exp, t.identifier("then")), [
     t.arrowFunctionExpression(
       [t.identifier("m")],
-      t.callExpression(file.addHelper("interopRequireWildcard"), [
+      t.callExpression(file.hub.addHelper("interopRequireWildcard"), [
```
pi0, line 24: (TODO: this is a runtime change need to check)

src/plugins/transform-module/index.ts
```diff
@@ -177,15 +177,14 @@ export default declare((api: any, options: Options) => {
     name: "transform-modules-commonjs",
 
     pre() {
-      // @ts-expect-error
-      this.file.set("@babel/plugin-transform-modules-*", "commonjs");
+      this.set("@babel/plugin-transform-modules-*", "commonjs");
 
-      // @ts-expect-error
-      if (lazy) defineCommonJSHook(this.file, lazyImportsHook(lazy));
+      if (lazy) defineCommonJSHook(this, lazyImportsHook(lazy));
```
pi0, line 182: (TODO: these are runtime changes. Was it essential?)

aryaemami59: I checked it against the source code from babel, the outcome is the same.

src/plugins/transform-module/dynamic-import.ts
aryaemami59: I checked it against the source code from babel, the outcome is the same.

aryaemami59: I can change these to `as` type assertions if you want.

src/plugins/transform-module/index.ts
aryaemami59: I reverted the runtime changes.

pi0 (approved):

PR #281, opened 2024-07-30
fix: use different cache dir as `jiti` v1

antfu:
As the transform results are different, in cases where a project uses both Jiti v1 and v2 together, the cache will interfere each others


pi0:
Both the extensions of entries (it is now `.cjs` or `.mjs` v1 was `.js`) are different and also jiti always has an integrity version. 

Did you happen to have any issues?

antfu:
Interesting, I wonder why I hit that. Let's close this for now and I will try to figure out an reproduction

PR #286, opened 2024-08-19
chore: replace exec logic in test fixtures

43081j:
Replaces `execa` with `tinyexec`, a much lighter alternative.

This should be equivalent since we only use the basics of either library.

(part of the ongoing ecosystem cleanup effort)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,8 @@
     "test:bun": "bun --bun test test/bun"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


Sec32fun32 (commented):
Yapılan araştırmalar sonucunda bu yayının incelenmesini talep ediyorum

pi0 (approved):

Issue #207, opened 2024-01-30
data-url modules fail to import


LorisSigrist:
### Environment

Node.js v21.2.0

### Reproduction

Self contained snippet: 

```js
import jiti from 'jiti'

const moduleAsText = "export default 42"
const dataUrl = "data:text/javascript;base64," + btoa(moduleAsText)

const nativeImport = await import(dataUrl)
console.log(nativeImport.default) // prints 42

const withJiti = jiti(import.meta.filename)(dataUrl)//Throws "MODULE_NOT_FOUND" error
console.log(withJiti.default)
```

### Describe the bug

`jiti` fails to import modules from data-urls, which is supported by node's native `import`

### Additional context

I tested both `base64` encodings and `text/javascript` encoding, neither worked.
Peering at the source-code there doesn't seem to be any handling for this yet. Would you be interested in supporting this?

We discovered this in opral/monorepo#2133, where a vite-plugin broke because of this.

### Logs

```sh
node:internal/modules/cjs/loader:1144
  const err = new Error(message);
              ^

Error: Cannot find module 'data:text/javascript;base64,ZXhwb3J0IGRlZmF1bHQgNDI='
Require stack:
- /Users/***/dev/jiti-debug/index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Function.resolve (node:internal/modules/helpers:187:19)
    at _resolve (/Users/***/dev/jiti-debug/node_modules/.pnpm/jiti@1.21.0/node_modules/jiti/dist/jiti.js:1:251148)
    at jiti (/Users/***/dev/jiti-debug/node_modules/.pnpm/jiti@1.21.0/node_modules/jiti/dist/jiti.js:1:253746)
    at file:///Users/***/dev/jiti-debug/index.js:11:13 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/***/dev/jiti-debug/index.js' ]
}
```
```

**Comment by bohjak** [comment]
As a temporary workaround, you can add an extension not in the extensions array[0] to the end of the URI.
This will make JITI think that it's an unsupported file type and will use the native import/require instead.

[0]: The default extensions array is: `[".js", ".mjs", ".cjs", ".ts", ".mts", ".cts", ".json"]`

### Example
```js
const script = `
export default 0;
// For JITI support when used as data: .unknown
`;
import(`data:text/javascript;charset=utf-8,${encodeURIComponent(script)}`);
```

PR #299, opened 2024-09-21
feat: handle `data:` imports (non-native only)

bohjak:
Resolves https://github.com/unjs/jiti/issues/207

Explicitly check for `data:` scheme and use native import to handle it.

### DIFF
--- a/src/require.ts
+++ b/src/require.ts
@@ -23,7 +23,7 @@ export function jitiRequire(
   } else if (id.startsWith("data:")) {
     if (opts.async === false) {
       throw new Error(
-        "data: URLs are not supported by CJS. Use import/jiti.import instead.",
+        "\`data:\` URLs are only supported in ESM context. Use `import` or `jiti.import` instead.",
       );
     } else {
       debug(ctx, "[native]", "[data]", "[import]", id);
--- a/src/require.ts
+++ b/src/require.ts
@@ -21,7 +21,7 @@ export function jitiRequire(
   } else if (id.startsWith("file:")) {
     id = fileURLToPath(id);
   } else if (id.startsWith("data:")) {
-    if (opts.async === false) {
+    if (!opts.async) {
       throw new Error(
         "\`data:\` URLs are only supported in ESM context. Use `import` or `jiti.import` instead.",
       );
--- a/src/require.ts
+++ b/src/require.ts
@@ -15,11 +15,14 @@ export function jitiRequire(
 ) {
   const cache = ctx.parentCache || {};
 
-  // Check for node: and file: protocol
+  // Check for node:, file:, and data: protocols
   if (id.startsWith("node:")) {
     id = id.slice(5);
   } else if (id.startsWith("file:")) {
     id = fileURLToPath(id);
+  } else if (id.startsWith("data:")) {
+    debug(ctx, "[native]", "[data]", opts.async ? "[import]" : "[require]", id);
+    return nativeImportOrRequire(ctx, id, opts.async);
   }
 
   // Check for builtin node module like fs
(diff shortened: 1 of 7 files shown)


src/require.ts
```diff
@@ -15,11 +15,14 @@ export function jitiRequire(
...
 
-  // Check for node: and file: protocol
+  // Check for node:, file:, and data: protocols
   if (id.startsWith("node:")) {
     id = id.slice(5);
   } else if (id.startsWith("file:")) {
     id = fileURLToPath(id);
+  } else if (id.startsWith("data:")) {
+    debug(ctx, "[native]", "[data]", opts.async ? "[import]" : "[require]", id);
+    return nativeImportOrRequire(ctx, id, opts.async);
```
pi0, line 25: I think this is only supported via async import (CJS does not support `data:`)

bohjak: oh! You're right. What should the behaviour be if `opts.async === true` then? Should we ignore it and always use the async `import`, error out, or something else?

pi0: Yes probably we can throw an error when async is false to say use import/jiti.import

bohjak: Done

src/require.ts
```diff
@@ -15,11 +15,20 @@ export function jitiRequire(
...
-  // Check for node: and file: protocol
+  // Check for node:, file:, and data: protocols
   if (id.startsWith("node:")) {
     id = id.slice(5);
   } else if (id.startsWith("file:")) {
     id = fileURLToPath(id);
+  } else if (id.startsWith("data:")) {
+    if (opts.async === false) {
+      throw new Error(
+        "data: URLs are not supported by CJS. Use import/jiti.import instead.",
```
pi0, line 26:
```suggestion
        "\`data:\` URLs are only supported in ESM context. Use `import` or `jiti.import` instead.",
```

src/require.ts
```diff
@@ -15,11 +15,20 @@ export function jitiRequire(
...
     id = fileURLToPath(id);
+  } else if (id.startsWith("data:")) {
+    if (opts.async === false) {
+      throw new Error(
+        "\`data:\` URLs are only supported in ESM context. Use `import` or `jiti.import` instead.",
+      );
+    } else {
+      debug(ctx, "[native]", "[data]", "[import]", id);
+      return nativeImportOrRequire(ctx, id, true);
+    }
```
pi0, line 31:
```suggestion
    if (opts.async === false) {
      throw new Error(
        "\`data:\` URLs are only supported in ESM context. Use `import` or `jiti.import` instead.",
      );
    }
    debug(ctx, "[native]", "[data]", "[import]", id);
    return nativeImportOrRequire(ctx, id, true);
```

src/require.ts
```diff
@@ -15,11 +15,19 @@ export function jitiRequire(
...
   const cache = ctx.parentCache || {};
 
-  // Check for node: and file: protocol
+  // Check for node:, file:, and data: protocols
   if (id.startsWith("node:")) {
     id = id.slice(5);
   } else if (id.startsWith("file:")) {
     id = fileURLToPath(id);
+  } else if (id.startsWith("data:")) {
+    if (opts.async === false) {
```
pi0, line 24:
```suggestion
    if (!opts.async) {
```

test/fixtures/data-uri/index.ts
```diff
@@ -0,0 +1,14 @@
...
+try {
+  require(uri);
+  throw null;
+} catch (error) {
+  assert(
+    error instanceof Error,
+    "Expected an error message for sync import of data URL",
+  );
+}
+import(uri);
```
pi0, line 14:
```suggestion
await import(uri);
```

bohjak:
@pi0 I've added the data-uri fixture to the ignore lists for the native and bun tests. Breaks native, because it requires ESM. I haven't looked into how exactly bun imports work, but for some reason the `require` keyword works fine with data URLs, but `import` doesn't :shrug:

pi0 (approved):
Thanks!

PR #305, opened 2024-09-26
fix: make `TransformOptions` type strict to allow auto-complete 

aryaemami59:
## **This PR**:

- [X] Fixes the `TransformOptions` type to not sacrifice auto-completion.

### DIFF
--- a/lib/types.d.ts
+++ b/lib/types.d.ts
@@ -177,9 +177,9 @@ export interface TransformOptions {
   ts?: boolean;
   retainLines?: boolean;
   interopDefault?: boolean;
-  async: boolean;
+  async?: boolean;
   jsx?: boolean | JSXOptions;
-  [key: string]: any;
+  babel?: Record<string, any>;
 }
 
 export interface TransformResult {


pi0:
Can you please elaborate where we loose auto complete?

aryaemami59:
Sure, currently if we have something like this:

```ts
createJiti(__filename, { transformOptions: {  } })
```

When we type inside `transformOptions` we get no autocompletion.

pi0:
Intresting seems it breaks `Omit<TransformOptions, "source">` usage!

aryaemami59:
> Intresting seems it breaks `Omit<TransformOptions, "source">` usage!

In what way?

pi0:
Actually [this line](https://github.com/unjs/jiti/blob/main/lib/types.d.ts#L182) is the problem. not sure why we had it before but removing has same effort (yet we can keep extendable interface consistent with other current types)

aryaemami59:
Yeah, that's why I removed that line, it was there before, not sure why. Does it look okay the way it is now? Or do you want me to make some other changes to it?

pi0:
I think removing that line should only be enough as it was tested locally so we can keep the interface mainly for consistency.

aryaemami59:
I get: `TS2339: Property 'babel' does not exist on type 'Omit<TransformOptions, "source">'.`, I'll add a `babel` property.

pi0 (commented):
Thanks!

PR #306, opened 2024-09-26
chore: update relative imports

aryaemami59:
## **This PR**:

- [X] Change the import path extension in `src/plugins/transform-module/lazy.ts` from `.ts` to `.js`.

Also not sure why `tsc` didn't catch this, I think it should since we don't have `allowImportingTsExtensions` turned on.

### DIFF
--- a/src/plugins/transform-module/lazy.ts
+++ b/src/plugins/transform-module/lazy.ts
@@ -3,7 +3,7 @@
 // https://github.com/babel/babel/tree/c7bb6e0f/packages/babel-plugin-transform-modules-commonjs/src
 import { template, types as t } from "@babel/core";
 import { isSideEffectImport } from "@babel/helper-module-transforms";
-import type { CommonJSHook } from "./hooks.ts";
+import type { CommonJSHook } from "./hooks.js";
 
 type Lazy = boolean | string[] | ((source: string) => boolean);
 


pi0:
I think we should turn `allowImportingTsExtensions` on instead and prefer real paths for imports.

aryaemami59:
It's up to you but it will probably be less unpredictable if we do things according to typeScript's `NodeNext` module resolution.

pi0:
Typescript made strange decisions over time. We follow ESM standards for imports (same as/compatible to Deno)

aryaemami59:
That's fair but can we at least do extensionless?

pi0:
Since jiti is not bundled i don't mind but would be better to update them with explicit extensions while we are at it so it will be compatible with Deno runtime and typescript host (i am migrating to explicit ES imports across other tools as well)

aryaemami59:
Okay, so basically We want to turn on `allowImportingTsExtensions` and update all relative import paths to have `.ts` extensions?

aryaemami59:
I get:

```bash
Option 'allowImportingTsExtensions' can only be used when either 'noEmit' or 'emitDeclarationOnly' is set.
```

pi0:
Yes we have to enable it as well. Here is a better tsconfig we should migrate to btw: https://github.com/unjs/rou3/blob/main/tsconfig.json

aryaemami59:
Looks good I'll use that!

aryaemami59:
If I turn on `allowImportingTsExtensions` with either `"noEmit": true` or `"emitDeclarationOnly": true`, the build step fails. I get the following error:

```bash
Error: TypeScript emitted no output for src\babel.ts.
Error: TypeScript emitted no output for src\jiti.ts.
```

pi0:
refactored extensionless for now.

PR #313, opened 2024-09-30
refactor: deprecate commonjs api

pi0:
Use `@deprecated` annotations to encourage migrating to ESM-compatible async API.


chadxz:
@pi0 I've been investigating esm-compatible synchronous require lately and was wondering - why did you decide to deprecate this? There are use-cases where being able to synchronously require a file is useful in node.js, such as loading configuration without requiring top-level await.

PR #320, opened 2024-10-05
docs: fix format

iiio2:


iiio2:
Thanks @pi0 . 
One thing, in About section in the repo, pls change it also. 

format

PR #321, opened 2024-10-05
fix(interopDefault): avoid `in` operator for primitive inputs

### LINKED ISSUE #319: TypeError: Cannot use 'in' operator to search for 'default' ...
### Environment

Jiti version 2.2.1 (the latest)
NodejS v22.8.0

### Reproduction

To reproduce, just install `npm install ffmpeg-static` (v5.2.0) and import it with Jiti like that:
```
import pathToFfmpeg from "ffmpeg-static";
```

This will throw this error every time:
```
TypeError: Cannot use 'in' operator to search for 'default' in /var/www/videomail/prod/source/node_modules/ffmpeg-static/ffmpeg
    at /var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:182495
    at jitiInteropDefault (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:183330)
    at jitiRequire (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:187464)
    at import (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:196818)
    at /var/www/videomail/prod/source/src/server/util/ffmpeg/runner.ts:3:50
    at eval_evalModule (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:193374)
    at jitiRequire (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:187618)
    at import (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:196818)
    at /var/www/videomail/prod/source/src/server/util/ffmpeg/encodeVideo.ts:8:44
    at async import (/var/www/videomail/prod/source/node_modules/jiti/dist/jiti.cjs:1:196812)
```

pi0:
resolves #319 (regression of #318)

If module export is a primitive like a number of string, `in` operator can throw an error. 

Instead, we can simply make sure `mod` is an object or (extended) function. (null/undefined default is still checked in line after)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -103,7 +103,7 @@ export function jitiInteropDefault(ctx: Context, mod: any) {
 }
 
 function interopDefault(mod: any): any {
-  if (!mod || !("default" in mod)) {
+  if (!mod) {
     return mod;
   }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -103,7 +103,8 @@ export function jitiInteropDefault(ctx: Context, mod: any) {
 }
 
 function interopDefault(mod: any): any {
-  if (!mod) {
+  const modType = typeof mod;
+  if (mod === null || (modType !== "object" && modType !== "function")) {
     return mod;
   }
 


src/utils.ts
```diff
@@ -103,7 +103,8 @@ export function jitiInteropDefault(ctx: Context, mod: any) {
 }
 
 function interopDefault(mod: any): any {
-  if (!mod || !("default" in mod)) {
+  const modType = typeof mod;
+  if (mod === null || (modType !== "object" && modType !== "function")) {
```
binarykitchen, line 107: What if mod is `undefined`? I'd add this check as well

src/utils.ts
```diff
@@ -103,7 +103,8 @@ export function jitiInteropDefault(ctx: Context, mod: any) {
 }
 
 function interopDefault(mod: any): any {
```
binarykitchen, line 105: `interopDefault` seems to be an important function. I'd add an extra comment explaining what it is about since the function name isn't self-explanatory. Thanks

src/utils.ts
pi0: `typeof undefined` is `undefined` (which passes the second part of the logic). only `null` is odd because `typeof null === 'object'`

src/utils.ts
pi0: PR welcome to improve readme explaining it better 👍🏼

src/utils.ts
binarykitchen: ACK

Issue #324, opened 2024-10-07
Regression when `require(extensionless-ts)` from JS


webpro:
### Environment

Regression when upgrading from Jiti v1 to Jiti v2.

Can reproduce in Node.js 18 and 22.

### Reproduction

See https://stackblitz.com/edit/stackblitz-starters-earn9d

Initially you can ignore `client` and `shared` folders, that holds the original dependency chain in the freeCodeCamp repo where Knip is used and `gatsby-node.js` is loaded with `jiti.import` (Jiti v2). I minimized that chain for a minimal repro to just two files: `index.js` and `req-ts.ts`.

Can see the issue here (`require(extensionless-ts)`) and maybe it wasn't supposed to work in the first place, but this loads without issues in Jiti v1:

```
$ npx jiti@^1 index.js
```

Yet with v2 this gives an error:

```
$ npx jiti@^2 index.js
Error: Cannot find module './req-ts'
Require stack:
- /Users/lars/p/knip/jiti-repro/index.js
```

### Describe the bug

With Jiti v1, Knip was loading this file without issues using the `jitiCJS` loader as created here: https://github.com/webpro-nl/knip/blob/b1cf45a9c82f16e83618e7841c2ec850d9b9d781/packages/knip/src/util/jiti.ts#L8-L21 (i.e. `esmResolve: false`, `extensions: []` and `interopDefault: true`).

I think it's a regression, but maybe I'm missing some config.

### Additional context

Didn't yet look into how in the original context Gatsby loads this, first wanted to check whether Jiti v2 is supposed to support this at all in the first place, or whether it's considered a breaking change.

pi0:
Hi dear @webpro thanks for spending time reproducing and explaining.

Using `JITI_DEBUG`, shows the situation:

<img width="663" alt="image" src="https://github.com/user-attachments/assets/8068b43e-a4d5-4987-b27a-942bfff0afa3">

The `index.js` passes all checks to be considered as a valid CommonJS module therefore jiti will try to import it natively.

It was the same in jiti v1 but jiti has a "fallback" mode that if prediction fails (native import fails), it tries again with transpile mode:

<img width="804" alt="image" src="https://github.com/user-attachments/assets/4ccd1246-28ad-48fc-9215-878dad7f24b1">

Jiti v2 also [has](https://github.com/unjs/jiti/blob/main/src/eval.ts#L72) the same fallback mode but it is not working in async context.

Looking for a solution.

pi0:
[...]

PR #326, opened 2024-10-07
fix(eval): return fallback value

webpro:
Resolve https://github.com/unjs/jiti/issues/324#issuecomment-2396804333.

Seems to do the trick. Let me if know if you'd need extra tests.


pi0 (approved):
Thanks!

(you can test with jiti-nightly after merge)

PR #328, opened 2024-10-07
Add pkg-pr-new workflow

webpro:
If you install this GitHub App, you'd get "nightlies" for free with any push or PR: https://github.com/stackblitz-labs/pkg.pr.new?tab=readme-ov-file#pkgprnew-

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -1,0 +1,21 @@
+name: Publish with pkg.pr.new
+
+on:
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR. I prefer to keep using npm for nightly channel as it requires no custom registry config today.

PR based previews are cool though and we are in contact to make it possible once pkg.pr.new is not depending on github internals.

webpro:
Fair enough! I love it, folks can just open their own PR + install anywhere.

PR #337, opened 2024-11-01
fix: try to resolve `.ts` files with `.js` extension from js files

### LINKED ISSUE #334: Importing ts file with `.js` specifier from `.mjs` in another workspace fails
### Environment

Node.js v20, both jiti v1 + v2

### Reproduction

Please see https://stackblitz.com/edit/stackblitz-starters-zwqrgl?file=playwright.config.mjs

```
npx jiti playwright.config.mjs 
```

Issue happens in both jiti v1 and v2. Cannot find module, but should print `{ hello: 1 }`

First I'd like to know if you think this should even work. If so, I'm happy to look into fix & send PR.

### Describe the bug

Not sure if this is even supposed to work: import with `.js` extensions in specifier which is actually a `.ts` file and in a workspace dir with `type: "module"`. Relevant bits (minimized in repro):

- https://github.com/argos-ci/argos/blob/main/playwright.config.mjs
- https://github.com/argos-ci/argos/blob/main/apps/backend/src/config/index.ts
- https://github.com/argos-ci/argos/blob/main/apps/backend/package.json#L5

Apparently Playwright's using something that's able to load it all up.

### Additional context

_No response_

### Logs

_No response_

webpro:
Resolves #334

Interestingly, with this change tests still pass and the use case is fixed. The change covers a regular use case I think: id's ending with JS resolve to a file with TS extension. BUT I'm probably missing something here, so please verify.

Btw, what's your dev mode to set breakpoints in source code directly (without build, register etc), is it possible? I often do something like `tsx --inspect src/cli.ts` but didn't succeed to create something like this.

### DIFF
--- a/src/resolve.ts
+++ b/src/resolve.ts
@@ -66,7 +66,8 @@ export function jitiResolve(
     // Try resolving .ts files with .js extension
     if (
       TS_EXT_RE.test(ctx.filename) ||
-      TS_EXT_RE.test(ctx.parentModule?.filename || "")
+      TS_EXT_RE.test(ctx.parentModule?.filename || "") ||
+      JS_EXT_RE.test(id)
     ) {
       resolved = tryNativeRequireResolve(
         ctx,


src/resolve.ts
```diff
@@ -63,11 +63,8 @@ export function jitiResolve(
     if (resolved) {
       return resolved;
     }
-    // Try resolving .ts files with .js extension
-    if (
-      TS_EXT_RE.test(ctx.filename) ||
-      TS_EXT_RE.test(ctx.parentModule?.filename || "")
```
pi0, line 69: While tests might pass (and I immediately cannot recall why we needed this check) it might be a safer option to add new checks in addition to the current ones.

webpro: Yeah just to make you aware of the situation. Pushed a minimal change/fix now.

Issue #57, opened 2022-03-25
Typescript: Add support for class properties and emitDecoratorMetadata


sreuss:
Hello,

switching Typescript transpilation to `jiti` in Nuxt 2.15.0 has caused some problems with our code base that have prevented us from upgrading:

We are using `class-validator` to validate data and define our classes like this:
```typescript
export class SomeModel {
  @IsString()
  @IsDefined()
  public readonly someField!: string;
}
```
This causes an `Error: TRANSFORM_ERROR: Definitely assigned fields cannot be initialized here, but only in the  constructor` on Nuxt startup which can be fixed by adding `@babel/plugin-proposal-class-properties` to the transformation. Somehow, `@babel/plugin-syntax-class-properties` is already included.

In addition, we are using Nest for our `serverMiddleware` and have some way to utilize it's dependency injection during server side rendering. Since the dependency injection relies on `emitDecoratorMetadata` we'd need `babel-plugin-transform-typescript-metadata` to be added to the transformation.

cshomo11:
I am also seeing issues trying to use `json2typescript` on a model in a Nuxt servermiddleware. I specifically see the error about using `proposal-class-properties`.

` ERROR  Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform.`

pi0:
Consdiering you already imported `IsString` and `IsDefined` this syntax works:

```js
export class SomeModel {
  @IsString()
  @IsDefined()
  public readonly someField: string;

  constructor(someField: string) {
    this.someField = someField;
  }
}
```

We also improved decorator meta support for v2.

octet-stream:
@pi0 I'm still having this problem on v2.4.2, so I don't think this solves the issue. It's pretty common to use the `!` operator in TS classes to mark properties that will be assigned later. For example: This is how you would define entities in Mikro ORM. See examples from their docs: https://mikro-orm.io/docs/guide/first-entity#first-entity

It works with TSC, works with SWC, and works with tsx, and this needs proper fix in jiti. This issue blocks me from integrating Mikro ORM with jiti, because it will cause too many problems for developers who use this library.

PR #361, opened 2025-01-30
fix: wrong babel plugin causes transform-class-properties error

Fgc17:
# Resolves

This PR resolves  [Originally posted by @cshomo11 in #57](https://github.com/unjs/jiti/issues/57#issuecomment-1335569520)

# MWE

[MWE (minimal reproduction) repository](https://github.com/Fgc17/jiti-transform-class-properties-error)

For testing the MWE, just hit `pnpm install` and then `pnpm working` or `pnpm not-working`.

# Description

We should not be using the `babel-plugin-syntax-class-properties`:

image

Just as written in the babel docs, the plugin does not bring transform, so it causes an issue with `class-transfomer` (and probably other packages):

```bash
[jiti] [init] version: 2.4.2 module-cache: true fs-cache: false interop-defaults: true
[jiti] [transpile] [esm] ./src/class-validator/env.config.ts (51.721ms)
[jiti] [native] [import] workspace/my-lib/packages/my-lib/dist/core/core.cjs
[jiti] [native] [import] workspace/my-lib/node_modules/.pnpm/dotenv@16.4.7/node_modules/dotenv/lib/main.js
[jiti] [native] [import] workspace/my-lib/node_modules/.pnpm/class-validator@0.14.1/node_modules/class-validator/cjs/index.js
[jiti] [native] [import] workspace/my-lib/node_modules/.pnpm/class-transformer@0.5.1/node_modules/class-transformer/cjs/index.js
workspace/my-lib/packages/playground/src/class-validator/env.config.ts:7
<TRANSPILED CODE THAT FAILED GOES HERE>

Error: Decorating class property failed. Please ensure that transform-class-properties is enabled and runs after the decorators transform.
    at _initializerWarningHelper (workspace/my-lib/packages/playground/src/class-validator/env.config.ts:7:767)
    at <instance_members_initializer> (workspace/my-lib/packages/playground/src/class-validator/env.config.ts:13:141)
    at new Schema (workspace/my-lib/packages/playground/src/class-validator/env.config.ts:13:116)
    at TransformOperationExecutor.transform (workspace/my-lib/node_modules/.pnpm/class-transformer@0.5.1/node_modules/class-transformer/cjs/TransformOperationExecutor.js:147:32)
    at ClassTransformer.plainToInstance (workspace/my-lib/node_modules/.pnpm/class-transformer@0.5.1/node_modules/class-transformer/cjs/ClassTransformer.js:27:25)
    at plainToInstance (workspace/my-lib/node_modules/.pnpm/class-transformer@0.5.1/node_modules/class-transformer/cjs/index.js:38:29)
    at validate (workspace/my-lib/packages/playground/src/class-validator/env.config.ts:27:60)
    at Command.<anonymous> (workspace/my-lib/packages/my-lib/dist/cli/cli.cjs:334:39)

Node.js v22.3.0
 ELIFECYCLE  Command failed with exit code 1.
```

The only fix without merging is including the plugin by yourself:
```ts
import pluginTransformClassProperties from "@babel/plugin-transform-class-properties";

const jiti = createJiti(import.meta.url, {
   transformOptions: {
     ts: true,
     babel: {
       plugins: [pluginTransformClassProperties],
     },
   },
 });
```

## Transpiled code that failed
```js
var _classTransformer = await jitiImport("class-transformer");
var _dec, _dec2, _dec3, _dec4, _class, _descriptor, _descriptor2;
function _interopRequireDefault(e) {
    return e && e.__esModule ? e : { default: e };
}
function _applyDecoratedDescriptor(i, e, r, n, l) {
    var a = {};
    return (
        Object.keys(n).forEach(function (i) {
            a[i] = n[i];
        }),
        (a.enumerable = !!a.enumerable),
        (a.configurable = !!a.configurable),
        ("value" in a || a.initializer) && (a.writable = !0),
        (a = r
            .slice()
            .reverse()
            .reduce(function (r, n) {
                return n(i, e, r) || r;
            }, a)),
        l && void 0 !== a.initializer && ((a.value = a.initializer ? a.initializer.call(l) : void 0), (a.initializer = void 0)),
        void 0 === a.initializer ? (Object.defineProperty(i, e, a), null) : a
    );
}
function _initializerWarningHelper(r, e) {
    throw Error("Decorating class property failed. Please ensure that transform-class-properties is enabled and runs after the decorators transform.");
}
let;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "@babel/helper-plugin-utils": "^7.25.9",
     "@babel/helper-simple-access": "^7.25.9",
     "@babel/plugin-proposal-decorators": "^7.25.9",
-    "@babel/plugin-syntax-class-properties": "^7.12.13",
+    "@babel/plugin-transform-class-properties": "7.25.9",
     "@babel/plugin-syntax-import-assertions": "^7.26.0",
     "@babel/plugin-syntax-jsx": "^7.25.9",
     "@babel/plugin-transform-export-namespace-from": "^7.25.9",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,14 +26,14 @@ importers:
       '@babel/plugin-proposal-decorators':
         specifier: ^7.25.9
         version: 7.25.9(@babel/core@7.26.0)
-      '@babel/plugin-syntax-class-properties':
-        specifier: ^7.12.13
-        version: 7.12.13(@babel/core@7.26.0)
       '@babel/plugin-syntax-import-assertions':
         specifier: ^7.26.0
         version: 7.26.0(@babel/core@7.26.0)
       '@babel/plugin-syntax-jsx':
         specifier: ^7.25.9
+        version: 7.25.9(@babel/core@7.26.0)
+      '@babel/plugin-transform-class-properties':
+        specifier: 7.25.9
         version: 7.25.9(@babel/core@7.26.0)
       '@babel/plugin-transform-export-namespace-from':
         specifier: ^7.25.9
@@ -163,13 +163,13 @@ importers:
         version: 3.8.0
       terser-webpack-plugin:
         specifier: ^5.3.11
-        version: 5.3.11(webpack@5.97.1)
+        version: 5.3.11(webpack@5.97.1(webpack-cli@5.1.4))
       tinyexec:
         specifier: ^0.3.1
         version: 0.3.1
       ts-loader:
         specifier: ^9.5.1
-        version: 9.5.1(typescript@5.7.2)(webpack@5.97.1)
+        version: 9.5.1(typescript@5.7.2)(webpack@5.97.1(webpack-cli@5.1.4))
       typescript:
         specifier: ^5.7.2
         version: 5.7.2
@@ -190,7 +190,7 @@ importers:
         version: 5.1.4(webpack-bundle-analyzer@4.10.2)(webpack@5.97.1)
       webpack-license-plugin:
         specifier: ^4.5.0
-        version: 4.5.0(webpack@5.97.1)
+        version: 4.5.0(webpack@5.97.1(webpack-cli@5.1.4))
       yoctocolors:
         specifier: ^2.1.1
         version: 2.1.1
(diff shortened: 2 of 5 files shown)


pi0:
Thank you for the PR and sorry for late review. I have added a dep reproduction and it fixes issue.

Testing on some projects, I found effect of this change is that babel is transpiling more than it should (for example private


```js
export class User {
  #id = 0;

  constructor() {
    this.#init();
    console.log(this.#id);
  }

  #init() {
    this.#id = Math.random();
  }
}
```

Was working before without any transform (JS has native private field/method support) but now fails:

> [Error: TRANSFORM_ERROR: Class private methods are not enabled. Please add `@babel/plugin-transform-private-methods` to your configuration.

Of course we can add more transforms but it is raising a question for me if it is safe thing to do.

One way we can do to move this faster is to make transform opt-in on top of syntax.

octet-stream:
> One way we can do to move this faster is to make transform opt-in on top of syntax.

Fine by me, but there might be more problems, so maybe things should be tested.

fgcoelho:
`plugin-transform-class-properties` only deals with classes, so every other transform we may need will probably be class-related. Therefore, I don't think we will need many more transforms.

Still, I support the idea of proceeding with opt-in transforms for now so we can address any arising issues and, after some time, merge them as enabled by default 👍

fgcoelho:
Hey guys 👋

I've implemented the "experimentalTransforms" option, along with a fixture for typestack packages.

You can test what happens without the option by going into `fixtures.test.ts` and setting `JITI_EXPERIMENTAL_TRANSFORMS: "false"`.

I've also fixed the missing MikroOrm fixture import in deps/index.ts

pi0:
Thank you! Perhaps better call it more explicitly `transformClassProps` ?

fgcoelho:
I would advocate for the `experimental` keyword if we were to make them default someday. We may need to add other transforms beyond the current two, so who knows if all `experimental` transforms will be related to transforming class properties?

Also, I don't think people should enable this option unless they got a real reason for it, like these edge cases with typestack and mikro-orm.

I may be wrong, let me know 🧐

pi0:
I think we would never enable this by default, as it implies more transformation and less proximity to runtime native behavior .

octet-stream:
> I think we would never enable this by default, as it implies more transformation and less proximity to runtime native behavior.

Does that mean that jiti's not aim to be compatible with tsc as much as possible? Because this specific feature should improve said compatibility.
I think if not, maybe it should be opt-in under more generic flag, of course if there's more things to improve there, aside from class transformations.

pi0:
Yes generic opt-in flag is cool 👍🏼 (my goal for jiti is to be as close to native Node.js as possible by default.)

octet-stream:
> my goal for jiti is to be as close to native Node.js as possible by default

So, `erasableSyntaxOnly` + `rewriteRelativeImportExtensions` + `allowImportingTsExtensions` compatibility by default, got it. Maybe it needs to be mentioned in readme somewhere? Because I don't really see it, but maybe it's just me :D

pi0:
Jiti supports all of those (TS features). I meant runtime (JS) syntax support like class props. It can be easily an opt in feature for when needed for class props :)

fgcoelho:
@pi0 Done. If it ain't going to be default anywhere in the future, then `transformClassProps` seems right.

fgcoelho:
@pi0 Done. If it ain't going to be default anywhere in the future, then transformClassProps seems right.

PR #377, opened 2025-05-05
fix: interop modules with nil default export

### LINKED ISSUE #376: TypeError: Cannot read properties of undefined ('configs') when importing eslint-plugin-react-compiler
### Environment

jiti version: 2.4.2
eslint-plugin-react-compiler version:19.1.0-rc.1

### Describe the bug

I've encountered a TypeError when trying to access the configs property of eslint-plugin-react-compiler when running a simple script with jiti. The same code works without issue when executed with Node.js.

### Reproduction

1. Create a file named `index.ts`.
2. Populate `index.ts` with the following code:

```
import reactCompiler from "eslint-plugin-react-compiler"

console.log(reactCompiler.configs.recommended)
```
3. Run the script using jiti:

```bash
npx jiti ./index.ts
```

### Expected Behavior

The script should successfully import eslint-plugin-react-compiler and log the recommended configuration object to the console, similar to how it behaves when run with Node.js.

### Actual Behavior

Running the script with jiti results in the following error:

```
TypeError: Cannot read properties of undefined (reading 'configs')
```

kricsleo:
resolves #376


pi0 (approved):
Thanks ❤️

Issue #383, opened 2025-06-09
Support custom user conditions in options


ernestostifano:
## TL;DR

It would be awesome if jiti could support [custom user conditions](https://nodejs.org/api/packages.html#resolving-user-conditions) in its options.

It would be something like:

`conditions`
- Type: Array
- Default: -
- Environment variable: JITI_CONDITIONS

Takes a list of additional [conditions](https://nodejs.org/api/packages.html#resolving-user-conditions) to be added to whatever existing conditions a resolver will use by default.

```js
const jiti = createJiti(import.meta.url, { conditions: ['development'] });
```

## Motivations

Thing is that setting `--conditions` in Node.js is not easy when third-party tools are involved. See point `5` at "The Problem - TL;DR" in this [Node.js Proposal](https://github.com/nodejs/node/issues/58626)

I really think jiti could easily make the difference while transitioning to a native alternative. It could allow achieving 100% ESM/TypeScript projects in a clean way, from configuration files to implementation code.

Also, I think we should really reconsider https://github.com/unjs/jiti/issues/34 in order to make options effective from within ESM files when we do not have control over how Node.js is executed:

```js
import "jiti/register";
```

`.jitirc`
```
JITI_CONDITIONS=['development']
```

I would be happy to help implementing this! Hope to get some feedback.

### Additional information

- [ ] Would you be willing to help implement this feature?

ernestostifano:
Would like to start working on a draft implementation and make some tests.

Non functional requirements on mind:
- Implementation does not introduce any breaking changes.
- Implementation does not introduce additional overhead or if minimum overhead is inevitable, it only applies when explicitly using the new feature.
- Interfaces remain consistent.
- Documentation is updated.
- Code follows project conventions, best practices and is clean/readable.

After looking at the source code, I was thinking that instead of implementing `.jitirc` idea, we could leverage the fact that we already access the "nearest `package.json`" file and we could just read an eventual `conditions` field from there and just pass it to the resolver by merging it with the context.

I known `conditions` is not a standard field (many aren't or became later), but it would satisfy the requirements and would be aligned with the Node.js proposal linked above.

What do you think @pi0? If you give me green light, I can start.

PR #384, opened 2025-06-13
feat: support custom user conditions

ernestostifano:
***Resolves:*** https://github.com/unjs/jiti/issues/383

### Main Changes:
- Added support for [custom user conditions](https://nodejs.org/api/packages.html#resolving-user-conditions) via `JitiOptions` and `JitiResolveOptions`, so, consistently across all APIs.
- Added support for [synchronous hooks](https://nodejs.org/api/module.html#synchronous-hooks-accepted-by-moduleregisterhooks).

### Added Capabilities

```js
const jiti = createJiti(import.meta.url, {conditions: true});
await jiti.import('some-package');
```

```js
const jiti = createJiti(import.meta.url, {conditions: ['development']});
await jiti.import('some-package');
```

```js
const jiti = createJiti(import.meta.url, {
    conditions: {'@scope/**/*': ['development']}
});
await jiti.import('some-package');
```

```js
const jiti = createJiti(import.meta.url, {
    conditions: [
        {
            match: ['@scope/**/*'],
            ignore: ['@scope/some-specific-package'],
            values: ['development']
        }
    ]
});
await jiti.import('some-package');
```

- There is full [minimatch](https://github.com/isaacs/minimatch) support.
- If `conditions` is set to `true`, configuration will be read from the nearest `package.json`, relative to `cwd`.

### Details:
- Created `utils.mjs` to group common code for both synchronous and asynchronous hooks.
- Replaced `exists` with `access` according to most recent [recommended approach](https://nodejs.org/api/fs.html#fsexistspath-callback).
- Added testing `pnpm` workspaces to enable testing with real packages and not only modules.
- Added tests for custom user conditions.
- Updated documentation.

### Notes
- There should not be any breaking changes.
- There should not be any relevant performance changes.
- Made `conditions` -> `true` by default to address some problems in this [Node.JS Proposal](https://github.com/nodejs/node/issues/58626) by being able to control `conditions` without controlling how Node.js is called (using `import 'jiti/regiter'`).
- I noted that there are some differences between `jiti` methods and `native` methods when using `import 'jiti/regiter'`. This is happening without this PR and I would be happy to investigate it and fix it, if needed, after merging this. Particularly:
    - `import()` output differs from `jiti.import()` output.
    - `createRequire()` -> `require()` output differs from `jiti.import()` output.
    - Static `import` output (when using `--import jiti/register` differs from `jiti.import()` output.


ernestostifano:
@pi0 happy to make any changes if needed.

ernestostifano:
@pi0 friendly ping. I'd love to receive sole feedback. Maybe I could help maintaining this package in the future.

If you think this cannot be merged right now I could publish a forked version.

pi0:
It is a risky changest with possible regressions and lots of changes. 

Ideally if someone can help making another PR with minimum enough (few lines of diff) to add user conditions it can speedup adding this feature.

PR #392, opened 2025-08-02
fix: always use native for `node:` specifiers

### LINKED ISSUE #391: Error: Cannot find module 'test'
### Environment

$ node -v
v22.17.0

jiti@2.5.1

### Reproduction

`test.mts`:
```typescript
import { test } from 'node:test';

test('noop', {});
```

Run with:
```shell
node --import jiti/register --test test.mts
```

Output:
```
node:internal/modules/cjs/loader:1401
  const err = new Error(message);
              ^

Error: Cannot find module 'test'
Require stack:
- <redacted>/test.mts
    at Function._resolveFilename (node:internal/modules/cjs/loader:1401:15)
    at Function.resolve (node:internal/modules/helpers:145:19)
    at jitiResolve (<redacted>/node_modules/jiti/dist/jiti.cjs:1:190064)
    at jitiRequire (<redacted>/node_modules/jiti/dist/jiti.cjs:1:192130)
    at import (<redacted>/node_modules/jiti/dist/jiti.cjs:1:202951)
    at _module (file://<redacted>/test.mts:1:259)
    at file://<redacted>/test.mts:10:7
    at ModuleJob.run (node:internal/modules/esm/module_job:329:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '<redacted>/test.mts' ]
}
```

### Describe the bug

When a script tries to import the `node:test` module, it crashes with the "Cannot find module 'test'" error.

I will try to submit a fix, but since I am not familiar with the codebase, it might take some time.

### Additional context

According to the [official documentation](https://nodejs.org/api/modules.html#built-in-modules-with-mandatory-node-prefix),

> Currently the built-in modules that requires the `node:` prefix are:
> 
> [node:sea](https://nodejs.org/api/single-executable-applications.html#single-executable-application-api)
> [node:sqlite](https://nodejs.org/api/sqlite.html)
> [node:test](https://nodejs.org/api/test.html)
> [node:test/reporters](https://nodejs.org/api/test.html#test-reporters)
>
> The list of these modules is exposed in [module.builtinModules](https://nodejs.org/api/module.html#modulebuiltinmodules), including the prefix.

*However*, according to the [history](https://nodejs.org/api/module.html#modulebuiltinmodules), prefix-only modules are included only since v23.5.0.

As far as I understand, the bug lives here: https://github.com/unjs/jiti/blob/61fa80358bc71708daf104c1205315d8e5867e4b/src/require.ts#L34-L36

The possible fix is to use [`module.isBuiltin()`](https://nodejs.org/api/module.html#moduleisbuiltinmodulename):

```
$ node
Welcome to Node.js v22.17.0.
Type ".help" for more information.
> const m = require('node:module');
undefined
> m.isBuiltin('node:test')
true
> m.isBuiltin('node:fs')
true
> m.isBuiltin('fs')
true
> m.isBuiltin('test') // should be false, as `test` is a prefix-only module
false
```

### Logs

```sh

```

sjinks:
resolves #391

This pull request refactors the `jitiRequire` function in `src/require.ts` to resolve the issue with detecting prefix-only built-in modules, such as `node:test`.

I'd love to add tests, but I was unable to find any that deal with the `node:` protocol.

To test:

`test.mts`:
```typescript
import { test } from 'node:test';

test('noop', {});
```

Run:
```sh
node --import jiti/register --test test.mts
```

Without this patch, a `MODULE_NOT_FOUND` error will occur.
With this patch, the script will successfully run.


src/require.ts
```diff
@@ -17,7 +17,7 @@ export function jitiRequire(
 
   // Check for node:, file:, and data: protocols
   if (id.startsWith("node:")) {
-    id = id.slice(5);
+    return nativeImportOrRequire(ctx, id, opts.async);
```
sjinks, line 20: When using `node:` protocol, `import`/`require` will fail on an unknown module:
```
node:internal/modules/esm/translators:358
    throw new ERR_UNKNOWN_BUILTIN_MODULE(url);
          ^

Error [ERR_UNKNOWN_BUILTIN_MODULE]: No such built-in module: node:x
    at ModuleLoader.builtinStrategy (node:internal/modules/esm/translators:358:11)
    at #translate (node:internal/modules/esm/loader:534:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27) {
  code: 'ERR_UNKNOWN_BUILTIN_MODULE'
}
```

That is why it probably does not make sense to run the `isBuiltin()` check: userland modules cannot use the `node:` protocol anyway.

src/require.ts
```diff
@@ -31,7 +31,7 @@ export function jitiRequire(
   }
 
   // Check for builtin node module like fs
-  if (builtinModules.includes(id) || id === ".pnp.js" /* #24 */) {
+  if (isBuiltin(id) || id === ".pnp.js" /* #24 */) {
```
pi0, line 34: The [`isBuiltIn`](https://nodejs.org/api/module.html#moduleisbuiltinmodulename) utility is available since Node.js >=16.17.0. We might need to be careful about backward compatibility and i guess with your fix above it might not be necessary anymore.

sjinks: `isBuiltin()` is [already used](https://github.com/search?q=repo%3Aunjs%2Fjiti%20isBuiltin&type=code) in the codebase.

PR #398, opened 2025-09-01
fix stack trace filename format for transpiled files

### LINKED ISSUE #397: Stack trace does not normalize windows paths
### Environment

windows

### Reproduction

https://github.com/kirkwaiblinger/repro-jiti-call-stack

### Describe the bug

I expect the stack trace to contain normalized windows paths that match node's format when a TS file executed by jiti is in the stack.

That means, CJS modules should have a `C:\path\to\file` format (resolved/normalized path) and ESM modules should have a `file:///C:/path/to/file` (file URL) format. 

Instead, jiti gives both cases as `C:/path/to/file`, which is neither resolved/normalized nor a valid file URL. 

### Additional context

Cause of https://github.com/typescript-eslint/typescript-eslint/issues/11530#issuecomment-3240528124

### Logs

```sh

```

kirkwaiblinger:
resolves #397 

## Summary by CodeRabbit

* **New Features**
  * Improved module-type detection and transpilation behavior: treat .mts as ESM, include .cts as CommonJS, always transpile TypeScript, and provide ESM runs with file-URL paths for accurate runtime reporting.

* **Tests**
  * Added end-to-end fixtures to validate filename/URL and stack-trace resolution across CJS, CTS, ESM, and MTS scenarios.


kirkwaiblinger:
@pi0 I've updated the tests. If you try out various combinations of reverting the changes to eval.ts and running the tests with windows/non-windows, it should hopefully demonstrate pretty clearly the things I'm trying to point out! 🙂

src/eval.ts
```diff
@@ -35,14 +37,17 @@ export function evalModule(
   const isTypescript = /\.[cm]?tsx?$/.test(ext);
   const isESM =
     ext === ".mjs" ||
-    (ext === ".js" && readNearestPackageJSON(filename)?.type === "module");
-  const isCommonJS = ext === ".cjs";
+    ext === ".mts" ||
+    ((ext === ".js" || ext === ".ts") &&
+      readNearestPackageJSON(filename)?.type === "module");
+  const isCommonJS = ext === ".cjs" || ext === ".cts";
   const needsTranspile =
```
kirkwaiblinger, line 44: btw this logic is obviously convoluted. It is equivalent to the following:

```ts
const needsTranspile = (() => {
  // force takes precedence
  if (evalOptions.forceTranspile != null) {
    return evalOptions.forceTranspile;
  }

  // TS always needs transpile, regardless whether ESM or CJS
  if (isTypescript) {
    return true;
  }

  // CommonJS skips transpile
  if (isCommonJS) {
    return false;
  }

  // In async mode, we can skip transpiling native ESM as well
  if (isESM && evalOptions.async) {
    return false;
  }

  return isESM || ctx.isTransformRe.test(filename) || hasESMSyntax(source);
})();
```

which I'd actually prefer to transcribe as 

```ts
const needsTranspile = (() => {
  // force takes precedence
  if (evalOptions.forceTranspile != null) {
    return evalOptions.forceTranspile;
  }

  // TS always needs transpile, regardless whether ESM or CJS
  if (isTypescript) {
    return true;
  }

  // CommonJS skips transpile
  if (isCommonJS) {
    return false;
  }

  
  if (isESM) {
    // In async mode, we can skip transpiling native ESM as well
    // otherwise it needs to be transformed
    return evalOptions.async;
  }

  // I don't know what this does
  if (ctx.isTransformRe.test(filename)) {
    return true;
  }

  if (hasESMSyntax(source)) {
     // Neither explicitly specified as ESM nor CJS, but contains 
     // ESM syntax. This should be transpiled.
     // See https://nodejs.org/api/packages.html#syntax-detection
     return true;
  }

  return false;
})();
```

The only reason I didn't do one of these is that the `prettier-ignore` made it seem like a previous author wanted the code to presented exactly as it is rather than in a more explicit manner.

The important logic change here is that `isTypescript` is no longer disjoint from `isESM` and `isCJS`. Distinguishing `isTypescript && isESM` from `isTypescript && isCJS` is necessary to present stack traces correctly for transpiled TS. Thus, now that `isTypescript && isCommonJS` is possible, the `isTypescript` check needed to be moved before the `isCommonJS`, else the `isCommonJS` check would cause `needsTranspile` to short-circuit to `false`.

kirkwaiblinger:
Looks like the test failure is due to bun formatting the filenames in stack traces differently (path rather than file URL) for ESM. Filed https://github.com/oven-sh/bun/issues/30298.

Regardless of whether the bun issue is accepted and/or fixed, I guess it wouldn't hurt to update the code/tests to be more lenient around whether a url vs a path is detected? Though one might need to have separate snapshot outputs for node vs bun vs deno then. Please advise 🙏

PR #403, opened 2025-09-19
perf: always use jiti/native in Deno

lucacasonato:
Deno already supports ESM/CJS interop, TypeScript, and tsx out of the box. It also doesn't have loader hooks for ESM anyway, so there is no point in trying to do that in Deno. Instead, we can always pass through to the native runtime functions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,10 @@
   "type": "module",
   "exports": {
     ".": {
+      "deno": {
+        "types": "./lib/jiti.d.mts",
+        "import": "./lib/jiti-native.mjs"
+      },
       "import": {
         "types": "./lib/jiti.d.mts",
         "default": "./lib/jiti.mjs"


pi0:
Thank you for the PR.  I see Deno native compat got much better 🚀  ([updated tests](https://github.com/unjs/jiti/commit/df844f8868500527f8cc46c1093147021c7b0a36#diff-2e4ab7a7b777191f5dbe7be66f5c8b665f6045489bf13b0742c6be8a75382711L11))

However, there are a couple of (non-standard) behaviors in jiti that are still not covered (you can test with `pnpm test:native:deno`). Because of this issue, we kept native mode opt-in (by consumer via explicit subpath).

We have another layer of fast-path to use the runtime native loader ([tryNative](https://github.com/unjs/jiti#trynative)). We enable it [by default for Bun](https://github.com/unjs/jiti/blob/df844f8868500527f8cc46c1093147021c7b0a36/src/options.ts#L28), I think it would be worth a shot to enable it also for Deno. It is 1) safe and 2) by the amount of native coverage, I think it should contribute to improving runtime perf. (I'm only worried if, in this case, Deno/Jiti users might start to see new warnings like implicit extensions)

lucacasonato:
That makes sense. Alternatively if you can lazy load things like babel only if they are actually used that may help too. Right now jiti is loaded by `@tailwind/node` unconditionally, and it contributes about 250ms to startup time from my measurements.

Issue #406, opened 2025-09-22
`TypeError` when accessing exported members inside namespaces in v2.6.0


Pixel998:
### Environment

Node v24.3.0
Jiti v2.5.1

### Reproduction

```ts
namespace MyNamespace {
  export const value = 123;
}

console.log(MyNamespace.value);
```

### Describe the bug

When trying to access any exported member inside a TypeScript namespace, I get the following error in v2.6.0:

```
TypeError: Cannot read properties of undefined (reading '<memberName>')
```

### Additional context

Downgrading to v2.5.1 resolves the issue.

### Logs

```sh

```

PR #407, opened 2025-09-30
build: revert to `terser-webpack-plugin`

kricsleo:
resolves https://github.com/unjs/jiti/issues/406

Revert from `rspack.SwcJsMinimizerRspackPlugin` back to `terser-webpack-plugin` due to a bundle issue. 
(This is a quick fix, more details need to be investigated)

As for the bundle size:

|`rspack.SwcJsMinimizerRspackPlugin`|`terser-webpack-plugin`|
|:-|:-|
|dist/jiti.cjs: `48,555 B`| dist/jiti.cjs: `48,382 B`|

### DIFF
--- a/rspack.config.mjs
+++ b/rspack.config.mjs
@@ -58,6 +58,8 @@ export default defineConfig({
     chunkIds: "named",
     minimize: true,
     minimizer: [
+      // TODO: Migrate to rspack.SwcJsMinimizerRspackPlugin
+      // https://github.com/unjs/jiti/pull/407
       new TerserPlugin({
         terserOptions: {
           mangle: {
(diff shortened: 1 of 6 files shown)


rspack.config.mjs
```diff
@@ -58,8 +58,8 @@ export default defineConfig({
     chunkIds: "named",
     minimize: true,
     minimizer: [
-      new rspack.SwcJsMinimizerRspackPlugin({
-        minimizerOptions: {
+      new TerserPlugin({
```
pi0, line 61:
```suggestion
      // TODO: Migrate to rspack.SwcJsMinimizerRspackPlugin
      // https://github.com/unjs/jiti/pull/407
      new TerserPlugin({
```

pi0 (approved):
Thanks!

PR #427, opened 2026-02-20
feat: support opt-in `tsconfigPaths` 

### LINKED ISSUE #373: Add feature flag to support typescript paths
### Describe the feature

Currently, jiti does not support [tsconfig.paths](https://www.typescriptlang.org/tsconfig/#paths), which significantly impacts DX. This limitation exists to ensure that [jiti remains portable and predictable by avoiding any reliance on external configuration files](https://github.com/unjs/jiti/issues/166#issuecomment-2374322954).

I’d like to propose adding a new option `tsconfigPaths/tsconfigAliases` which would be set to `false` by default. When enabled, jiti would resolve imports based on [tsconfig.paths](https://www.typescriptlang.org/tsconfig/#paths). This approach preserves the default behavior and predictability of jiti, while offering an opt-in improvement for those who want better DX in TypeScript projects.

The [original comment](https://github.com/unjs/jiti/issues/166#issuecomment-2785349904) where the proposal was discussed.

### Additional information

- [ ] Would you be willing to help implement this feature?

rexxars:
**Resolves TypeScript `compilerOptions.paths` aliases at runtime, matching tsc behavior.**

Resolves #373 

## Usage

```ts
const jiti = createJiti(import.meta.url, { tsconfigPaths: true });
```

- `true` - auto-discover `tsconfig.json` by walking up from the jiti instance's parent path
- `string` - explicit path to a `tsconfig.json`
- `false` (default) - disabled (backwards compatible)

Also configurable via `JITI_TSCONFIG_PATHS` environment variable.

## Implementation

Uses [get-tsconfig](https://www.npmjs.com/package/get-tsconfig) (same library used by _tsx_ and _bundle-require_) to parse the tsconfig and create a paths matcher. The matcher is created once per jiti instance, not re-parsed on every resolve call.

In the resolve chain (`resolve.ts`), tsconfig path candidates are tried before alias resolution and before Node's native resolution. Each candidate is resolved through jiti's existing file resolution (including extension handling), and the
first that exists on disk wins.

Handles the full tsc path matching spec:

- Wildcard patterns (`@/*` → `./src/*`)
- Exact matches (`@config` → `./src/config`)
- Catch-all (`*` → `./types/*`)
- Mid-pattern wildcards (`foo/*/bar` → `./src/*/baz`)
- Multiple fallback targets (`@/*` → `./src/*`, `./generated/*`) - tries each in order
- extends chains and baseUrl resolution (handled by `get-tsconfig`)

Edge cases:
- No tsconfig found → silent no-op
- No paths in tsconfig → skip resolution
- Both tsconfigPaths and alias set → tsconfig paths run first, then alias as a second pass

## Aside

Really happy with jiti - it handles compatibility better than both tsx and bundle-require - the only thing missing is TS paths support :)

Let me know if you need something to change! I ran into an issue when running tests locally that I attempted to fix in https://github.com/unjs/jiti/pull/426

## Summary by CodeRabbit

* **New Features**
  * Added a public option to enable TypeScript path-alias resolution (boolean or string); supports auto-discovery or explicit tsconfig path and respects JITI_TSCONFIG_PATHS.
* **Documentation**
  * README updated with usage, types, and environment variable details.
* **Types**
  * Public types updated to include the new option and related context capability.
* **Tests**
  * Added fixture demonstrating resolving modules via tsconfig path aliases.


pi0 (approved):
Thanks ❤️

rexxars:
Thanks for merging! Any chance of a new release anytime soon? Hope to release something depending on this and would rather not use a fork :)

pi0:
surely soon. please ping me by this week if forgot.

rexxars:
@pi0 Ping as requested ;)

rexxars:
Hate to keep pinging, but wondering if there's any chance of a new release @pi0 ?

jsardev:
Would be great to have this released 🙏 Can't wait!

pi0:
Hi. Sorry for delay it is out! https://github.com/unjs/jiti/releases/tag/v2.7.0

PR #430, opened 2026-02-27
feat: add `jiti/static` export

### LINKED ISSUE #417: Make `babel` transform lazy loading optional
## Context

Jiti's lazy-loading approach for the Babel transform is the best option for most use cases. Since the Babel transform requires loading a substantial dependency, deferring its load until needed provides meaningful performance benefits when the transform isn't required. However, I think I ran into a use case where it might make sense to eagerly load it.

We work with a runtime environment that **only supports CommonJS** execution. The workflow involves:
   1. Writing source code in ESM for better developer experience
   2. Using `jiti` to dynamically load configuration files at runtime
   3. Bundling everything with Webpack into a single CJS file for deployment

```ts

async function loadConfig(name) {
  const jiti = createJiti(__filename);
  const path = resolve(__dirname, `./config/config.${name}`);

  const config = await jiti.import(path, { default: true });
  return config;
}
```

## Problem

When bundling with Webpack, the ESM version of `jiti` (`dist/lib/jiti.mjs`) contains this code:

```ts
let _transform;
function lazyTransform(...args) {
  if (!_transform) {
    _transform = createRequire(import.meta.url)("../dist/babel.cjs");
  }
  return _transform(...args);
}
```

And Webpack's default behavior is to replace import.meta.url with a static file path at build time:

```ts
// After Webpack processes it
_transform = createRequire("file:///Users/username/project/node_modules/jiti/lib/jiti.mjs")("../dist/babel.cjs");
```

This hardcoded absolute path doesn't exist in the production runtime, causing the lazy load to fail:

```
Error: Cannot find module '../dist/babel.cjs'
```

## Workaround

I managed to workaround it by using the CJS distributed version via `require('jiti')` as it does not use `import.meta.url` under the hood: 

```ts
const { createJiti } = require("jiti");  // Uses lib/jiti.cjs
```
Webpack can properly handle and bundle this require() call. However, the CommonJS entry point is marked as deprecated, and I'm concerned about future compatibility.

## Proposed solutions

Of course, if someone has a better solution please let me know. From my point of view I think either of these two options below could work:

### Option 1: Configuration Flag

```ts
const jiti = createJiti(__filename, {
  eagerTransform: true  // Loads babel immediately
});
```

### Option 2: Separate Entry Point
```ts
import { createJiti } from "jiti/eager";
```

### Additional information

- [x] Would you be willing to help implement this feature?

pi0:
Adds a new `jiti/static` export that statically imports `babel.cjs` instead of lazy-loading it via `createRequire(import.meta.url)`.

Closes #417

This solves two use cases:
- **Bun compiled binaries**: ensures `babel.cjs` gets bundled into the binary
- **Webpack/bundler compatibility**: the lazy `createRequire(import.meta.url)` pattern breaks when bundlers rewrite `import.meta.url` to a hardcoded path. A static import lets bundlers resolve and include `babel.cjs` correctly.

Usage:
```js
import { createJiti } from "jiti/static";
```

Based on https://github.com/badlogic/jiti/commit/800a4fc547273afa40a9c7ca1d7e0166542f8ce8

Co-authored-by: Mario Zechner <badlogicgames@gmail.com>

### DIFF
--- a/lib/jiti-static.mjs
+++ b/lib/jiti-static.mjs
@@ -1,0 +1,23 @@
+import { createRequire } from "node:module";
+import _createJiti from "../dist/jiti.cjs";
+// Static import so Bun bundles babel.cjs into compiled binaries
+import _babelTransform from "../dist/babel.cjs";
+
+function onError(err) {
+  throw err; /* ↓ Check stack trace ↓ */
+}
+
+const nativeImport = (id) => import(id);
+
+export function createJiti(id, opts = {}) {
+  if (!opts.transform) {
+    opts = { ...opts, transform: _babelTransform };
+  }
+  return _createJiti(id, opts, {
+    onError,
+    nativeImport,
+    createRequire,
+  });
+}
+
+export default createJiti;
--- a/package.json
+++ b/package.json
@@ -24,6 +24,10 @@
       "types": "./lib/jiti.d.mts",
       "import": "./lib/jiti-native.mjs"
     },
+    "./static": {
+      "types": "./lib/jiti.d.mts",
+      "import": "./lib/jiti-static.mjs"
+    },
     "./package.json": "./package.json"
   },
   "main": "./lib/jiti.cjs",
@@ -35,6 +39,9 @@
         "./lib/jiti-register.d.mts"
       ],
       "native": [
+        "./lib/jiti.d.mts"
+      ],
+      "static": [
         "./lib/jiti.d.mts"
       ]
     }


RaphaelDDL:
Hi @pi0 , can we have a release for this please?

jiti going from 2.5.1 to 2.6.1 (e.g. from nuxt/kit, that comes with nuxt/test-utils) when running `npm audit fix` to fix vulnerabilities, bumps jiti thus makes running Nuxt break with the `Cannot find module '../dist/babel.cjs'` error.

_(btw, I'm not using bun nor webpack, it's just default nuxt with vite and nuxi so not even sure if that new import will fix, since I can't change nuxt/kit way of import)_

iivvaannxx:
@pi0 sorry for the ping, would like to have this released too please 🙏🏻

PR #435, opened 2026-04-10
fix(require): normalize windows paths for native import when tryNative is true

medns:
## Summary
- Fixes `ERR_UNSUPPORTED_ESM_URL_SCHEME` on Windows when `tryNative` is true and Jiti falls back to transpilation.
- During fallback, if a module contains a dynamic `import()`, Jiti transforms it to `jiti.import()`, which internally calls `ctx.nativeImport()` with `opts.async = true`.
- On Windows, Node's native `import()` requires absolute paths to be `file://` URLs. This PR uses the existing `normalizeWindowsImportId` helper to convert paths before passing them to `nativeImport()`.

## Test plan
- Ran `pnpm test` successfully.
- Verified that all 28 fixtures pass.
- Ensured no regressions in native or CJS interop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured Windows import path normalization is applied consistently for native async imports, matching other import paths and preventing import failures.
  * Normalization utility now consistently returns a string for absolute Windows paths, removing edge-case type inconsistencies.

### DIFF
--- a/src/require.ts
+++ b/src/require.ts
@@ -59,7 +59,7 @@ export function jitiRequire(
       );
       if (opts.async && ctx.nativeImport) {
         return ctx
-          .nativeImport(id)
+          .nativeImport(normalizeWindowsImportId(id))
           .then((m: any) => {
             if (ctx.opts.moduleCache === false) {
               delete ctx.nativeRequire.cache[id];
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -164,11 +164,11 @@ function interopDefault(mod: any): any {
   });
 }
 
-export function normalizeWindowsImportId(id: string) {
+export function normalizeWindowsImportId(id: string): string {
   if (!isWindows || !isAbsolute(id)) {
     return id;
   }
-  return pathToFileURL(id);
+  return pathToFileURL(id).toString();
 }
 
 let _fipsMode: boolean | undefined;


pi0:
Thanks for PR. I have low confidence since there is no relavant issue or minimal reproduction. Can you please provide it?

medns:
Thanks for reviewing!

The issue occurs on Windows when using Jiti with `tryNative: true` to load an ESM module that contains a dynamic `import()`.

When a native `require()` fails on an ESM module, Jiti falls back to transpilation. During this process, Jiti transforms any dynamic `import()` within the module to `jiti.import()`, which internally calls `ctx.nativeImport()` with `opts.async = true`. If the path is an absolute Windows drive-letter path (e.g. `C:\path\to\module`), Node's native `import()` will throw an `ERR_UNSUPPORTED_ESM_URL_SCHEME` error because it requires absolute paths to be formatted as `file://` URLs.

This was explicitly reported and encountered in the wild by the `openclaw` project. They had to implement a workaround in their own code to convert paths to `file://` URLs before passing them to Jiti (see: https://github.com/openclaw/openclaw/issues/61911 and their workaround `toSafeImportPath`).

Here is a minimal reproduction to demonstrate the issue:

```javascript
// dep.mjs
export const a = 1;
```

```javascript
// entry.mjs
export async function load() {
  // Dynamic import triggers the issue
  return import("D:\\jiti\\dep.mjs");
}
```

```javascript
// test-repro.cjs
const { createJiti } = require("jiti");

const jiti = createJiti(__filename, { tryNative: true });

try {
  // Synchronous load triggers the fallback because entry.mjs is ESM
  const mod = jiti("D:\\jiti\\entry.mjs");
  console.log("Loaded entry synchronously!");
  
  // Calling the function triggers the dynamic import -> jiti.import -> ctx.nativeImport
  mod.load().then(
    () => console.log("Loaded dep successfully!"),
    (err) => console.error("Failed to load dep:", err)
  );
} catch (e) {
  console.error("Failed to load entry:", e);
}
```

Without this PR, running `node --no-experimental-require-module test-repro.cjs` on Windows outputs:

```
Loaded entry synchronously!
Failed to load dep: Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'd:'
...
```

With this PR, Jiti correctly handles the path conversion internally using `normalizeWindowsImportId` in the `tryNative` path, resolving the issue.

PR #454, opened 2026-07-22
fix: respect Node preserve-symlinks resolution

### LINKED ISSUE #445: [Bug]: jiti config import resolver in rspack fails to respect --perserve-symlinks / jiti gives no symlink-following customization options to fix this
### Environment

System Info
System:
OS: Linux 6.8 Ubuntu 22.04.5 LTS 22.04.5 LTS (Jammy Jellyfish)
CPU: (8) x64 Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz
Memory: 37.98 GB / 61.80 GB
Container: Yes
Shell: 5.1.16 - /usr/bin/bash
Binaries:
Node: 20.19.5 - ~/.nvm/versions/node/v20.19.5/bin/node
Yarn: 4.12.0 - ~/src/server/node_modules/.bin/yarn
npm: 10.8.2 - ~/.nvm/versions/node/v20.19.5/bin/npm
Watchman: 20240414.112832.0 - /usr/local/bin/watchman
npmPackages:
@rspack/cli: ^2.0.0 => 2.0.0
@rspack/core: ^2.0.0 => 2.0.0
@rspack/plugin-react-refresh: ^2.0.0 => 2.0.0

## Reproduce link
https://github.com/dgoldstein0/rspack-jiti-symlink-bug-repro

## Reproduce Steps
follow the readme in my repro repo.

The key ingredients:

invoke rspack via node --preserve-symlinks --preserve-symlinks-main node_modules/@rspack/cli/bin/rspack.js - so that we have some expectation that rspack will not follow symlinks when resolving the config's imports.
at least one import in the rspack config (and perhaps the config itself) should point to symlinks.
the symlink points outside the folder where node_modules are installed. so if the symlink is followed, node_modules imported in the symlinked file won't be found. (alternatively, relative imports wouldn't work either)
the imported symlink must be of a form that the native node module loader won't resolve and rspack will fall back to using jiti (unless rspack is passed --loader jiti). This is easy to achieve by leaving off the file extension for the import in an ESM rspack config.

### Describe the bug

When resolving imports within a config file that don't resolve under built in node rules, rspack falls back to the jiti resolver which seems to follow symlinks whether or not --perserve-symlinks is passed to the underlying node. For my use case this resulted in it finding the next file at the symlink target, and then failing to find transitive imports - as the transitive imports were discoverable from the symlink, not the file it targeted.  From what I can see jiti has no options to customize symlink handling, so this is arguably not rspack's fault.

I discovered this while wrangling issues updating to rspack 2 & converting to an ESM rspack config in my company's bazel integration; bazel heavily uses symlinks and I got a similar "node module can't be found" by rspack config import resolution escaping the tree where my node_modules were symlinked.

### Additional context

_No response_

### Logs

```sh

```

rajanpanth:
## Summary

- retain Node's symlink-preserved resolution path when Jiti and the native resolver select the same underlying file
- preserve import-condition behavior by rejecting native candidates that canonicalize to a different file
- add a cross-platform regression for an extensionless symlinked TypeScript helper with a transitive package dependency

Fixes #445.

## Problem

When preserve-symlinks behavior is enabled, native Node resolution can retain the logical symlinked path while mlly returns the physical realpath.

Jiti previously returned mlly's canonicalized path before its native fallback could preserve the symlinked path. This caused the extensionless TypeScript helper to execute from its physical location, where a dependency available only through the package root could not be resolved.

## Root cause

mlly canonicalizes a successful resolution using `realpathSync`. Jiti accepted that result before reconciling it with the native resolution path used under preserve-symlinks behavior.

## Change

Under preserve-symlinks mode, compare mlly's resolved file with the native fallback.

The native path is retained only when both paths canonicalize to the same underlying file. If import conditions or other resolution semantics select a different file, mlly's result remains unchanged.

## Validation

- focused preserve-symlinks regression — 1/1 passed
- Vitest with coverage — 36 passed, 4 skipped
- Node-register tests — 29/29 passed
- native Node tests — 15/15 passed
- typecheck — passed
- lint — passed
- targeted formatting — passed
- direct Rspack build — passed
- `git show --check` — passed

Validated on Windows 11 with Node 24.18.0 and repository-pinned pnpm 10.30.3.

Node 20, Node 22, Linux, and macOS were not tested locally.

The aggregate `pnpm build` command could not complete its `rm -rf dist` cleanup because the Unix-style `rm` command is unavailable in this PowerShell environment. The direct Rspack build completed successfully.

## Test configuration

`vitest.config.ts` only adds `test/preserve-symlinks.test.ts` to the repository's explicit test include list. It does not alter timeouts, exclusions, coverage thresholds, production behavior, or unrelated test scope.

## Compatibility

No public API, dependency, lockfile, TypeScript declaration, or supported Node-version change is included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module resolution when Node.js symlink preservation is enabled.
  * Ensured imports from symlinked paths resolve to the expected modules without errors.
  * Added support for detecting symlink-preservation settings from command-line options and environment configuration.
  * Improved fallback handling for modules resolved with extensions or directory index files.

* **Tests**
  * Added coverage for symlink-preserving resolution, including temporary package layouts and Node.js preservation flags.

### DIFF
--- a/src/resolve.ts
+++ b/src/resolve.ts
@@ -101,10 +101,9 @@ function tryNativeFallback(
   parentURL: URL | string,
   options: { paths?: string[] },
 ) {
-  try {
-    return ctx.nativeRequire.resolve(id, { paths: options.paths });
-  } catch {
-    // Try additional extensions below.
+  const nativeResolved = tryNativeRequireResolve(ctx, id, parentURL, options);
+  if (nativeResolved) {
+    return nativeResolved;
   }
 
   let resolved;
--- a/src/resolve.ts
+++ b/src/resolve.ts
@@ -1,3 +1,4 @@
+import { realpathSync } from "node:fs";
 import { resolveAlias } from "pathe/utils";
 import { fileURLToPath, resolvePathSync } from "mlly";
 import { join, dirname } from "pathe";
@@ -6,6 +7,11 @@ import { isDir } from "./utils";
 
 const JS_EXT_RE = /\.(c|m)?j(sx?)$/;
 const TS_EXT_RE = /\.(c|m)?t(sx?)$/;
+const PRESERVE_SYMLINKS =
+  process.execArgv.includes("--preserve-symlinks") ||
+  /(?:^|\s|["'])--preserve-symlinks(?:$|\s|["'])/.test(
+    process.env.NODE_OPTIONS || "",
+  );
 
 export function jitiResolve(
   ctx: Context,
@@ -65,16 +71,43 @@ export function jitiResolve(
       lastError = error;
     }
     if (resolved) {
-      return resolved;
+      return preserveSymlinkPath(ctx, id, resolved, parentURL, options);
     }
   }
 
   // Try native require resolve with additional extensions and /index as fallback
+  resolved = tryNativeFallback(ctx, id, parentURL, options);
+  if (resolved) {
+    return resolved;
+  }
+
   try {
-    return ctx.nativeRequire.resolve(id, { paths: options.paths });
+    ctx.nativeRequire.resolve(id, { paths: options.paths });
   } catch (error) {
     lastError = error;
   }
+
+  if (options?.try) {
+    // Well-typed in types.d.ts
+    return undefined as unknown as string;
+  }
+
+  throw lastError;
+}
+
+function tryNativeFallback(
+  ctx: Context,
+  id: string,
+  parentURL: URL | string,
+  options: { paths?: string[] },
+) {
+  try {
+    return ctx.nativeRequire.resolve(id, { paths: options.paths });
+  } catch {
+    // Try additional extensions below.
+  }
+
+  let resolved;
   for (const ext of ctx.additionalExts) {
     resolved =
       tryNativeRequireResolve(ctx, id + ext, parentURL, options) ||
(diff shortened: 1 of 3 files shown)


rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to update anything if needed.

Thanks!

src/resolve.ts
```diff
@@ -65,16 +71,42 @@ export function jitiResolve(
...
 
   // Try native require resolve with additional extensions and /index as fallback
+  resolved = tryNativeFallback(ctx, id, parentURL, options);
+  if (resolved) {
+    return resolved;
+  }
+
   try {
-    return ctx.nativeRequire.resolve(id, { paths: options.paths });
+    ctx.nativeRequire.resolve(id, { paths: options.paths });
```
official-burak: This used to `return` the native resolve result. After the split it still calls `nativeRequire.resolve`, then ignores success and falls through to `throw lastError`.

`tryNativeFallback` is not the same call: it prepends `dirname(parentURL)` to `paths`. This one uses `options.paths` as-is. A specifier that only resolves that way now fails, and a successful resolve here is discarded.

Restore the `return`, or delete the dead try/catch.

official-burak (commented):
Native resolve success is dropped after the helper split. Specifiers that only resolve via `options.paths` (no parent dir prepended) now fail.

rajanpanth:
Good catch, you're right on both counts.

The `return` was dropped when I split the fallback into `tryNativeFallback`. As written the call resolved and then discarded the result, falling through to `throw lastError`.

And the two calls aren't interchangeable. `tryNativeRequireResolve` passes `paths: [dirname(fileURLToPath(parentURL)), ...options.paths]`, and it also swallows any throw from `fileURLToPath`, so a specifier that only resolves from `options.paths` as-is genuinely had no path through the new code.

Restored the `return` in 7a1313f rather than deleting the block, since it's reachable for exactly that case.

Verified: vitest 37 passed / 4 skipped, `tsgo --noEmit` clean, eslint clean. The one failure I see locally is `hmr` in `test/bun`, which fails identically on this branch with the change reverted, so it looks environmental on my side rather than related.

official-burak:
Thanks. `tryNativeFallback` now returns the `paths`-only resolve result instead of discarding it. That was the hole.

PR #456, opened 2026-08-11
fix(tsconfig): resolve paths from referenced projects in solution-style tsconfigs

### LINKED ISSUE #440: tsconfigPaths behaves not the same as tsc
### Environment

Node.js v22.22.0 - jiti@latest

### Reproduction

https://stackblitz.com/edit/vitejs-vite-vzxdmpru?file=tsconfig.json

### Describe the bug

- split tsconfig.json in seperate referenced configs
- define paths in tsconfig.node.json not in the main tsconfig.json
- include test.ts file which includes src/test.ts by '@/test'
- npm run test # tsc -b && JITI_TSCONFIG_PATHS=1 npx -y jiti ./test.ts
- tsc works - jiti not

Workaround:
- in main tsconfig.json add  "compilerOptions": { "paths": { "@/*": ["./src/*"] } }
- now it works in tsc and jiti

Seems to be, that jiti not support references with paths.

### Additional context

_No response_

### Logs

```sh

```

pacocartones:
### 🔧 Fix: resolve `tsconfigPaths` from referenced projects in solution-style tsconfigs

Fixes #440 — `tsconfigPaths` does not behave like `tsc -b` when a solution-style `tsconfig.json` keeps its `paths` in a referenced project (e.g. `tsconfig.node.json`).

**Repro:** a root `tsconfig.json` containing only `"files": []` + `"references": [{ "path": "./tsconfig.node.json" }]`, with `paths: { "@/*": ["./src/*"] }` defined in `tsconfig.node.json`. `tsc -b` resolves `@/test`; `jiti` with `tsconfigPaths: true` fails with `MODULE_NOT_FOUND: Cannot find module '@/test'` because `getTsconfig()` returns the root config, which has no `paths`, and `createPathsMatcher()` only looks at the config that was found.

**Fix (`src/jiti.ts`):** when the found tsconfig produces no paths matcher, `createReferencedPathsMatcher()` walks its `references` (breadth-first, cycle-safe, with the base directory tracked per project so nested references in subdirectories resolve relative to the config that declares them) and composes one `createPathsMatcher()` per referenced project that declares `paths`. The fallback only activates when the root config has no `paths` — existing behavior is untouched.

### Test (red → green, executed)

New fixture `test/fixtures/tsconfig-references/` reproduces the issue: solution-style root tsconfig, `@/*` alias in the referenced project, `main.ts` imports `@/test`.

| State | Commit | Result |
|---|---|---|
| Red | `c567045` (fixture only) | `Cannot find module '@/test'` — the exact error from #440 |
| Green | `43cc2e8` (fix) | fixture passes, snapshot `"meaning of life: 42"` |

The `main.ts` fixture is added to the `tsconfig.json` `exclude` (same pattern the repo already uses for `test/fixtures/error-parse/index.ts`), since the alias cannot resolve at type level in the root project — that is precisely the bug.

### Checks

- `pnpm build` (rspack) ✅
- `pnpm lint` (eslint + prettier) ✅
- `pnpm test:types` (tsgo --noEmit) ✅
- `pnpm vitest run` — 36/36 passed, 4 skipped, no regressions ✅

## Summary by CodeRabbit

- **New Features**
  - Added support for resolving TypeScript path aliases defined in referenced project configurations.
  - Alias resolution now works when the root TypeScript configuration delegates settings to another configuration.
  - Supports nested project references and safely handles circular references.

- **Bug Fixes**
  - Improved module loading for projects using solution-style TypeScript configurations.

- **Tests**
  - Added coverage for path alias resolution across referenced, nested, and circular TypeScript configurations.

### DIFF
--- a/src/jiti.ts
+++ b/src/jiti.ts
@@ -6,9 +6,10 @@ import type {
   EvalModuleOptions,
   JitiResolveOptions,
 } from "./types";
+import { existsSync, statSync } from "node:fs";
 import { platform } from "node:os";
 import { fileURLToPath, pathToFileURL } from "mlly";
-import { join, dirname } from "pathe";
+import { join, dirname, basename, resolve } from "pathe";
 import escapeStringRegexp from "escape-string-regexp";
 import { normalizeAliases } from "pathe/utils";
 import pkg from "../package.json";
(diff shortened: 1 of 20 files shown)


pacocartones:
Done — added both suggested fixtures in `4fc801d` (verified green locally):

- **`tsconfig-references-nested`**: solution root → `packages/app` → `packages/lib`, with `paths` in the deepest config. Exercises recursive traversal and per-project base dirs (`@lib/*` resolves relative to `packages/lib`, not the root).
- **`tsconfig-references-cycle`**: root → `a` → `b` → `a` (cycle back to an already-visited config). Proves the traversal terminates via the visited set instead of looping.

Full run on the branch: build ✅ · lint ✅ · test:types ✅ · vitest 38/38 (incl. the two new fixtures).

pacocartones:
This adds a `tsconfigPaths` fallback that walks `references` when a solution-style root tsconfig has no `paths` of its own (fixes #440), with fixtures covering the direct, nested and cyclic cases. I noticed main hasn't had a commit since the v2.7.0 release in May and there are 22 PRs open, so rather than nudge I'd rather just ask honestly: is jiti still taking outside contributions at the moment? If it isn't, or this isn't a direction you'd want, I'm glad to close it rather than leave it sitting in the queue.

PR #460, opened 2026-08-17
fix(babel): allow Stage 3 decorator transforms

### LINKED ISSUE #381: Cannot use with Stage 3 Decorators
### Environment

`@better-auth/cli -> c12 -> jiti`
Node 22.14.0

### Reproduction

1. Create a file with a Stage 3 decorator that trying to access context.kind (second argument of a decorator)
2. Wrap some class with it
3. Try to load config with it, receive error `TypeError: Cannot read properties of undefined (reading 'kind')`
4. Try to add Babel plugin into transformOptions.babel.plugins, try again. Observe the same error

### Describe the bug

My project uses latest Stage 3 Decorators proposal in our codebase, and we found that `c12` uses jiti to load config. We tried to pass `babel.plugins`:  ["@babel/plugin-proposal-decorators", { "version": "2023-11" }], but without any luck. It ends on the end of a list, so previous transformations already tried to convert code with legacy decorators.

### Additional context

I've tried to load the config with just jiti, this is how I found that the error is in jiti setup. Patching c12 or @better-auth/cli doesn't provide any benefit

### Logs

```sh

```

yhuikzdtguioaert:
Fixes #381

## What changed

- detect a caller-provided `@babel/plugin-proposal-decorators` configuration and avoid injecting Jiti's legacy decorator transform alongside it
- resolve the package-name form to Jiti's bundled decorator plugin, so both Babel's string and function plugin forms work
- retain the existing legacy decorator behavior when no custom decorator transform is configured
- add regression coverage that executes transformed Stage 3 decorator output and verifies `context.kind`

## Testing

- `pnpm exec eslint src/babel.ts test/transform.test.ts vitest.config.ts`
- `pnpm exec prettier -c src lib test stubs`
- `pnpm test:types`
- `pnpm vitest run --coverage` (37 passed)
- `pnpm rspack`
- Node register/native suites (29 + 15 passed)
- Bun native suite (25 passed)
- Deno native suite (22 passed)

Codex assisted with repository exploration, implementation, tests, and validation. The resulting behavior and diff were reviewed before submission.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Stage 3 decorators by package name or plugin object.
  * Preserved decorator plugin options during transformation.
  * Decorators now run correctly before TypeScript processing, including decorated fields and field initialization.
* **Bug Fixes**
  * Ensured custom decorator configurations take precedence over legacy defaults.
* **Tests**
  * Expanded coverage for decorator execution, plugin formats, and decorator context handling.

### DIFF
--- a/src/babel.ts
+++ b/src/babel.ts
@@ -42,6 +42,12 @@ export default function transform(opts: TransformOptions): TransformResult {
         normalizeProposalDecoratorsPlugin(plugin),
       )
     : [];
+  const customDecoratorsPlugins = customPlugins.filter((plugin) =>
+    isProposalDecoratorsPlugin(plugin),
+  );
+  const otherCustomPlugins = customPlugins.filter(
+    (plugin) => !isProposalDecoratorsPlugin(plugin),
+  );
   const _opts: BabelTransformOptions & { plugins: PluginItem[] } = {
     babelrc: false,
     configFile: false,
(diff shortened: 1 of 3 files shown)


yhuikzdtguioaert:
Good catch on the ordering — fixed.

The two sources quoted in the review contradict each other, so worth being explicit about which one applies: the Babel documentation for `@babel/plugin-proposal-decorators` puts the decorator transform **before** `@babel/plugin-transform-typescript`. That is also what jiti already does — the built-in legacy decorators plugin is `unshift`ed, so it runs ahead of the TypeScript transform.

That is exactly what my patch broke. Caller plugins were appended at the end of the list, so a caller-supplied decorators plugin ended up running *after* `@babel/plugin-transform-typescript`, i.e. in a different position from the default it replaces. The opt-in worked, but it silently changed the plugin order.

Now a caller-supplied decorators plugin takes the same slot the legacy default occupied instead of being appended:

* `customPlugins` is split into decorator plugins and everything else;
* when `opts.ts` is set, the decorator plugins are unshifted alongside `transformTypeScriptMetaPlugin`, and only the remaining custom plugins are appended;
* when `opts.ts` is not set there is no TypeScript transform to order against, so custom plugins keep their previous position at the end — no behaviour change there;
* multiple decorator plugins keep the caller's relative order.

Test coverage extended for both reference forms (the imported plugin and the `"@babel/plugin-proposal-decorators"` package name) with a class that combines a `declare` field and a decorated typed field, so the assertion actually depends on the transforms running in the right order rather than only on the plugin being picked up.

